PR #46, opened 2025-07-22
Implement JWT-based authentication carrying the usecase/scope

Swatinem:
This implements authentication in the following way:
- All the HTTP APIs expect an `Authorization` header carrying a JWT.
- This JWT will carry the `usecase` and `scope`.
- Those two, along with the (optionally auto-assigned) `key` form the final `ObjectKey`.
- The Service backend can use this `ObjectKey` to make further decisions on where/how to store the actual object.

---

REF FS-63


objectstore-service/src/metadata.rs
```diff
@@ -0,0 +1,46 @@
...
+///
+/// While the storage service offers organization and project scopes, it does not maintain any
+/// association between those two, and it is up to the user to provide all projects belonging to an
+/// organization.
+#[derive(Debug, Deserialize, PartialEq, Eq)]
+#[serde(rename_all = "snake_case")]
+pub enum Scope {
+    Organization(u64),
+    Project(u64),
+}
```
jan-auer, line 28: +1 on making this an enum so we can extend this in the future. A few questions around this:

1. Should the project scope also include the organization? For example, if I save an object for a project and then query it with organization-wide authorization, I should also be able to read it. For this, the object needs to know which organization it belongs to. Potentially, this needs to become a struct with optional fields, instead?
2. In addition to the scope with ID, should we also expose an enumeration for the kind of scope ("organization" vs "project")?
3. This serializes using the default external tagging as `{"organization": 4711}`. Would we consider custom serialization, such as `"org:4711"` to use this easily in string-only places like request headers?

objectstore-service/src/metadata.rs
```diff
@@ -0,0 +1,46 @@
...
+
+/// The per-object expiration policy
+///
+/// We support time-to-live and time-to-idle policies.
+#[derive(Debug, PartialEq, Eq)]
+pub enum ExpirationPolicy {
```
jan-auer, line 34: I assume this is planning to have `Option<ExpirationPolicy>` at the place where this is referenced. As a thought, should we instead include a `Never` or `NoExipration` variant and implement `Default`?

objectstore-service/src/metadata.rs
```diff
@@ -0,0 +1,46 @@
+use serde::Deserialize;
+use std::time::Duration;
+
+/// Denotes the storage usecase
+///
+/// Different decisions can be made based on this usecase, such as using a different underlying
+/// storage backend.
+#[derive(Debug, Deserialize, PartialEq, Eq)]
+#[serde(rename_all = "snake_case")]
+pub enum Usecase {
```
jan-auer, line 10:
```suggestion
pub enum UseCase {
```

objectstore-service/src/metadata.rs
```diff
@@ -0,0 +1,46 @@
...
+    Attachments,
```
jan-auer, line 11: Let us make `UseCase` wrap a plain String - if we care about `Copy` we can swap this out with an interned string. 

We need to allow produce use cases to be added without code changes of our service or ideally the client library. Ideally, use cases are pure configuration supplied to our service at startup. If we want to type use cases in the client, I would suggest we expose this from sentry_protos or derive a type from the place where the use cases are configured.

Note that this doesn't have to be pure runtime-config, our service should just be agnostic over _specific_ use cases.

objectstore-server/src/http.rs
```diff
@@ -53,39 +51,52 @@ struct PutBlobResponse {
     key: String,
 }
 
+impl Claim {
```
jan-auer, line 54: _nit_: Since this is defined in a different module, can we move the impl block?

objectstore-server/src/authentication.rs
```diff
@@ -0,0 +1,94 @@
...
+use serde::Deserialize;
+
+#[derive(Debug, Deserialize, PartialEq, Eq)]
+pub struct Claim {
+    pub usecase: Usecase,
+    pub scope: Scope,
+}
+
+pub fn decode_auth_header(token: &str) -> JwtResult<Claim> {
+    let key = DecodingKey::from_secret(b"TODO");
```
jan-auer, line 16: Let's load the key from config? We can then define it using environment variables.

PR #50, opened 2025-07-25
test: Add an integration test using stresstest as a library

jan-auer:
Adds an integration test to the `objectstore-server` crate that spawns the full
server in a subprocess and then runs the stresstest against it for 10 seconds.

To do so, the `stresstest` crate has been split into a library and binary:
- Workload definition and the stresstest itself are in the library and can be
  called via `stresstest::run`.
- YAML configuration and a main function that reads config, configures
  workloads, and then runs the stresstest remain in the binary.

The initial stresstest integration test is not meant to mimic any real workload
and it does not run in release mode. Rather, it is used to integration-test the
stresstest with the binary. In a follow-up, I want to use `objectstore-sdk` in
stresstest, to ensure the client can talk to the server.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -471,6 +471,12 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "9555578bc9e57714c812a1f84e4fc5b4d21fcb063490c624de019f7464c91268"
 
 [[package]]
+name = "cfg_aliases"
+version = "0.2.1"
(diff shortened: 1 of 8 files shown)


Swatinem (approved):
> spawns the full server in a subprocess

Having such an "end to end" test surely is valuable, as it also validates the graceful shutdown, etc.

Alternatively, I would rather start the server in-process on an auto-assigned port, and run tests against that.

PR #55, opened 2025-08-05
Revise the way how we implement compression

Swatinem:
This also splits the client into multiple files, so it better organizes the code for PUT/GET.

I hope I understood our discussion related to compression correctly.

On PUT, you can supply an explicit compression algorithm, and on GET, you can tell the client to not automatically decompress, but rather return compressed data.


objectstore-client/src/client.rs
```diff
@@ -0,0 +1,166 @@
...
+            .build()?;
+        let jwt_key = EncodingKey::from_secret(jwt_secret.as_bytes());
+
+        Ok(Self {
+            service_url: service_url.trim_end_matches('/').into(),
+            client,
+
+            jwt_key,
+            usecase: usecase.into(),
+            default_compression: Compression::Zstd,
```
matt-codecov, line 64: i think it would be good to make this default a global const with a doc comment

```
/// Changing this requires a major version bump.
///
/// If a user updates this crate infrequently, then a single update may introduce a new compression algorithm
/// _and_ make that algorithm the default. This can complicate deployments; the first host to get the update
/// will use a compression algorithm that the other hosts don't understand yet.
pub(crate) const DEFAULT_COMPRESSION: Compression = Compression::Zstd;

impl ClientBuilder {
  pub fn new(...) {
    Self {
      default_compression: DEFAULT_COMPRESSION,
      ...
    }
  }
}
```

PR #78, opened 2025-09-09
Add a second storage backend

Swatinem:
This hardcodes two storage backends, one for small files, and one for large files. The threshold has been chosen arbitrarily as 50 KiB for now.

### DIFF
--- a/objectstore-server/src/config.rs
+++ b/objectstore-server/src/config.rs
@@ -31,7 +31,8 @@ pub struct Config {
     pub http_addr: SocketAddr,
 
     // storage config
-    pub storage: Storage,
+    pub small_storage: Storage,
+    pub large_storage: Storage,
 
     // authentication config
     pub jwt_secret: String,
@@ -45,7 +46,10 @@ impl Default for Config {
         Self {
             http_addr: "0.0.0.0:8888".parse().unwrap(),
 
-            storage: Storage::FileSystem {
+            small_storage: Storage::FileSystem {
+                path: PathBuf::from("data"),
+            },
+            large_storage: Storage::FileSystem {
                 path: PathBuf::from("data"),
             },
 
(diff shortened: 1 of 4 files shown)


objectstore-server/src/config.rs
```diff
@@ -31,7 +31,8 @@ pub struct Config {
     pub http_addr: SocketAddr,
 
     // storage config
-    pub storage: Storage,
+    pub small_storage: Storage,
+    pub large_storage: Storage,
```
jan-auer, line 35: Can we give these storages different designation than "small" and "large" in config? I think this conveys wrong semantics. 

From our cost modeling, we've seen that the two primary reasons we need two backends are:
- Retention time: With long retention (release files and debug files), resident data becomes the cost driver, prompting a backend with cold storage options.
- QPS: With high operation throughput, the per-ops cost becomes the cost driver, prompting a backend with low or no per-ops cost, at the expense of more costly storage.

The file size is an initially sufficient heuristic to make the decision of which backend to use, but we should decouple this from how the backends are identified. I can quickly think of "high_volume" and "long_term", what would you suggest?

objectstore-service/src/lib.rs
```diff
@@ -73,16 +74,33 @@ impl StorageService {
...
-        let key = ObjectKey::for_backend(1);
+        let mut first_chunk = BytesMut::new();
+        let mut backend_id = 1; // 1 = small files backend
+        while let Some(chunk) = stream.try_next().await? {
+            first_chunk.extend_from_slice(&chunk);
+
+            if first_chunk.len() > SMALL_THRESHOLD {
+                backend_id = 2; // 2 = large files backend
+                break;
+            }
```
jan-auer, line 87: We need to ensure in the interface that overwrites of _existing_ objects in the same backend are possible. I will finish the [decision on object mutability](https://www.notion.so/sentry/Are-objects-mutable-25d8b10e4b5d8019b92cf30c9acc5b47?source=copy_link) today - expect that we will go with option 3 at the bottom:

> Allow payload updates with the same ID and actually replace the data (or use versioning in the backend where available). If the new payload exceeds hard limits of the backend that the object was previously stored in, we reject the update.

This implies that we also need to add a second hard limit for when to reject the update. Example with real values:
- Threshold for BigTable -> GCS is 1MB. 
- Hard limit for BigTable is 10MB.

objectstore-service/src/lib.rs
```diff
@@ -91,15 +109,37 @@ impl StorageService {
         &self,
         key: &ScopedKey,
     ) -> anyhow::Result<Option<(Metadata, BackendStream)>> {
-        self.0.backend.get_object(key).await
+        match key.key.backend {
+            1 => self.0.small_backend.get_object(key).await,
+            2 => self.0.large_backend.get_object(key).await,
+            _ => anyhow::bail!("invalid backend"),
+        }
```
jan-auer, line 116: _nit_: Can we refactor this into a helper function and introduce consts for the magic numbers?

objectstore-service/src/lib.rs
```diff
@@ -18,17 +20,21 @@ use crate::backend::{BackendStream, BoxedBackend};
 pub use backend::BigTableConfig;
 pub use metadata::*;
 
+/// The threshold up until which we will go to the small backend.
+const SMALL_THRESHOLD: usize = 50 * 1024; // 50 KiB
```
jan-auer, line 24: Let's set this at 1MB right away, all backends we're evaluating are efficient up to that size, and this covers our use cases well.

jan-auer (changes requested):

Swatinem:
1. The current API does not allow for overwrites at all currently, as I changed it to not allow for an ID on `put`.
2. You would have to add a different backend with another id, and keep the existing config as-is as long as all existing files expire.

PR #88, opened 2025-09-15
ref(service): Cleanup backend naming and config

jan-auer:
Changes the BigTable and GCS implementations to align in their configuration and initialization: 

 - No more configuration via environment variables, the endpoint can always be passed in
 - If an endpoint is specified, we assume an emulator and skip authentication via gcp_auth
 - Configuration is passed as borrowed constructor arguments instead of a config struct
 
This also re-enables a higher-level test for the GCS backend, since this is available in CI now.

Ref FS-119


Swatinem (approved):

objectstore-service/src/lib.rs
```diff
@@ -55,12 +54,25 @@ pub enum StorageConfig<'a> {
     /// Use Google Cloud Storage as storage backend.
     Gcs {
         /// Optional endpoint URL for the S3-compatible storage.
+        ///
+        /// Assumes an emulator without authentication if set.
```
matt-codecov, line 58: perhaps we should make this its own config, or infer it based on whether GCP project / credentials are configured

PR #93, opened 2025-09-17
feat(client): propagate sentry-trace header

matt-codecov:
adds an `enable_distributed_tracing` option which, if on, will populate the `sentry-trace` header on outgoing requests to objectstore.

is it worth extra boilerplate to avoid coupling our client library with `sentry-rust`? the "correct" way to do this is to externalize the dependency and take an `Option<Fn() -> Option<(String, String)>>` which is expected to return a `(header_name, trace_id)` pair. and then i guess we give users a copy-pastable snippet or separate crate for plugging in the specifics for Sentry.


objectstore-client/src/client.rs
```diff
@@ -115,25 +131,53 @@ impl fmt::Debug for Client {
...
+        if self.enable_distributed_tracing {
+            // If the Sentry SDK has been initialized and tracing is in progress,
+            // this will set the header Sentry uses to propagate distributed
+            // tracing context. Otherwise, nothing happens.
+            if let Some(trace_id) = sentry::configure_scope(|scope| {
+                scope
+                    .get_span()
+                    .map(|span| span.get_trace_context().trace_id)
+            }) {
+                builder = builder.header(SENTRY_TRACE_HEADER, trace_id.to_string());
```
Swatinem, line 168: There is [`Scope::iter_trace_propagation_headers`](https://docs.rs/sentry/latest/sentry/struct.Scope.html#method.iter_trace_propagation_headers) for this purpose, so you don’t need to add a hardcoded header, but the SDK will expose any headers in whichever format it wants, in a forward compatible way.

objectstore-client/src/client.rs
```diff
@@ -26,6 +28,7 @@ impl fmt::Debug for ClientBuilder {
             .field("client", &self.client)
             .field("usecase", &self.usecase)
             .field("default_compression", &self.default_compression)
+            .field("propagate_traces", &self.propagate_traces)
```
Swatinem, line 31: now that we have no secret anymore that needs "protecting" in the `Debug` impl, might as well replace this with a derive, as that is the only thing that is left here.

objectstore-client/src/client.rs
```diff
@@ -115,6 +129,7 @@ impl fmt::Debug for Client {
             .field("usecase", &self.usecase)
             .field("scope", &self.scope)
             .field("default_compression", &self.default_compression)
+            .field("propagate_traces", &self.propagate_traces)
```
Swatinem, line 132: same here

Swatinem (approved):

PR #95, opened 2025-09-18
fix: Add Secret newtype to avoid config secrets in logs

matt-codecov:
adds a `Secret` newtype around `String` which:
- does not implement `Display`
- writes `"[redacted]"` in its `Debug` implementation
- allows access to the underlying secret as a `&str` via `secret.access()`

we already do `debug!(?config)` and inevitably we will add things like `debug!(?request.headers); // Temporary log to debug #1234`. use this type when you have the opportunity for values you recognize are sensitive so that logging like this doesn't cause an incident

the sentry dsn and datadog key are not "defcon 0" secrets but they are definitely non-public and we will probably have more sensitive examples in the future

### DIFF
--- a/objectstore-server/src/config.rs
+++ b/objectstore-server/src/config.rs
@@ -1,13 +1,48 @@
 use std::collections::BTreeMap;
+use std::fmt;
 use std::net::SocketAddr;
 use std::path::PathBuf;
 
 use anyhow::Result;
 use argh::FromArgs;
 use figment::providers::{Env, Format, Serialized, Yaml};
+use secrecy::{CloneableSecret, SecretBox, SerializableSecret, zeroize::Zeroize};
 use serde::{Deserialize, Serialize};
 
 const ENV_PREFIX: &str = "FSS_";
+
+/// Newtype around `String` that may protect against accidental
+/// logging of secrets in our configuration struct. Use with
+/// [`secrecy::SecretBox`].
+#[derive(Clone, Default, Serialize, Deserialize, PartialEq)]
+pub(crate) struct ConfigSecret(String);
+
+impl ConfigSecret {
+    pub fn as_str(&self) -> &str {
+        self.0.as_str()
+    }
+}
+
+impl std::ops::Deref for ConfigSecret {
+    type Target = str;
+    fn deref(&self) -> &Self::Target {
+        &self.0
+    }
+}
+
+impl fmt::Debug for ConfigSecret {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> Result<(), fmt::Error> {
+        write!(f, "[redacted]")
+    }
+}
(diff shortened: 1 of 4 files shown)


Swatinem (approved):
lgtm, good idea.

there are a bunch of publishes crates that go even further and zero out the memory on `Drop`, and maybe pulling in a dependency for this might be a good idea?

objectstore-types/src/lib.rs
```diff
@@ -203,3 +203,36 @@ impl Metadata {
         Ok(headers)
     }
 }
+
+/// Newtype around [`String`] that can protect against certain cases of
+/// logging sensitive values. Please use this instead of `String` in
+/// things like configuration structs which may be dumped blindly into
+/// debug logs.
+#[derive(Deserialize, Clone, Serialize)]
```
jan-auer, line 211: As an idea, what if we also implement `Display` like so, to ensure it can never be accidentally added?

```
impl fmt::Display for InternalId {
    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
        if cfg!(debug_assertions) {
            unimplemented!("do not display secrets");
        } else {
            write!(f, "[redacted]")
        }
    }
}
```

jan-auer (approved):

objectstore-types/src/lib.rs
Swatinem: You surely mean "implement it as a panic" so that noone will implement it any other way?
Because otherwise, by not implementing it at all, the compiler and type system will make sure noone will accidentally add a *usage* of the trait.

jan-auer: > so that noone will implement it any other way

Exactly, that's the idea. It does come with the unfortunate downside that now Display is in fact implemented. Maybe it's enough to trust future reviewers, though.

objectstore-server/src/config.rs
```diff
@@ -1,14 +1,49 @@
...
+
+impl ConfigSecret {
+    pub fn as_str(&self) -> &str {
+        self.0.as_str()
+    }
+}
+
+impl std::ops::Deref for ConfigSecret {
+    type Target = String;
+    fn deref(&self) -> &Self::Target {
```
Swatinem, line 28: `&String` is a pretty useless type, usually you deref to `&str`.

Swatinem (approved):
not opposed to this, but I honestly expected things to get better when relying on a crate, and not worse. pulling in `secrecy` adds a shitton of complexity without any real benefit IMO

objectstore-server/src/config.rs
matt-codecov: oh i can probably change this back to `str`. i changed it to `String` because i thought build errors were telling me to, but the fix was something else

PR #109, opened 2025-09-24
ci: enable ci workflows to run on merge_group trigger

matt-codecov:
enabling merge queues before doing this means it will wait forever for required workflows that are never triggered


Swatinem (approved):
thanks for fixing my oversight

PR #205, opened 2025-11-15
feat: implement S3 compatible API with authentication support

### LINKED ISSUE #127: Support AWS v4 authentication for S3-compatible backend
https://docs.aws.amazon.com/AmazonS3/latest/API/sig-v4-authenticating-requests.html

Should support this since most S3-compatible backend that self-hosted users uses requires a certain authentication using at least "S3 access token" + "S3 secret token". We can defaults the region name to `auto`.

aldy505:
Closes #127

### DIFF
--- a/objectstore-server/src/config.rs
+++ b/objectstore-server/src/config.rs
@@ -150,7 +150,7 @@ pub enum Storage {
         /// Or for long-term storage:
         /// - `OS__LONG_TERM_STORAGE__TYPE=s3compatible`
         /// - `OS__LONG_TERM_STORAGE__ENDPOINT=https://s3.amazonaws.com`
-        endpoint: String,
+        endpoint: Option<String>,
 
         /// S3 bucket name.
         ///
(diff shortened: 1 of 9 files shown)


objectstore-server/src/config.rs
```diff
@@ -161,6 +161,62 @@ pub enum Storage {
...
+        /// - `OS__LONG_TERM_STORAGE__SESSION_TOKEN=my-session-token`
+        session_token: Option<String>,
+
+        /// Optional request timeout for the S3-compatible storage.
+        ///
+        /// # Environment Variables
+        ///
+        /// - `OS__HIGH_VOLUME_STORAGE__REQUEST_TIMEOUT_SECS=30`
+        /// - `OS__LONG_TERM_STORAGE__REQUEST_TIMEOUT_SECS=30`
+        request_timeout_secs: Option<u64>,
```
Swatinem: this should probably be a `Duration`, with a `humantime` annotation.

objectstore-service/src/backend/s3_compatible.rs
```diff
@@ -1,131 +1,117 @@
...
-        Ok(())
+        Self { bucket }
+    }
+
+    async fn handle_get_object_response(
+        &self,
+        _path: &ObjectPath,
+        response: ResponseData,
+    ) -> Result<Option<(Metadata, BackendStream)>> {
+        let metadata = Metadata::from_hashmap(response.headers(), "").unwrap_or_default();
```
Swatinem: this should propagate any error that happens.

objectstore-service/src/backend/s3_compatible.rs
```diff
@@ -1,131 +1,117 @@
...
+        let bytes = Bytes::from(response.to_vec());
+        let stream: BackendStream = Box::pin(stream::iter(std::iter::once(Ok(bytes))));
```
Swatinem: S3 will most likely be used for large payloads, and we do not want to buffer those in memory. There should be some kind of way to get a stream from the `ResponseData`.

aldy505: There is actually: https://docs.rs/rust-s3/latest/s3/bucket/struct.Bucket.html#method.get_object_stream

But it's only compatible with async File IO type. I'm struggling to convert it into `reqwest`'s streaming format.

lcian:
Let's close this for now to maintain a clear list of open PRs. Many things have changed, so there's many merge conflicts right now. We might be able to reuse this or parts of this in the future, when we have a clearer idea on how to proceed with S3 support.

PR #419, opened 2026-04-02
feat(py-client): Add many API

jan-auer:
Adds `session.many()` to the Python client for executing multiple get/put/delete operations in batches, with a fully streaming request/response pipeline.

The batch protocol uses multipart/form-data with per-part custom headers (`x-sn-batch-operation-kind`, `x-sn-batch-operation-key`, etc.). No existing Python library supports both streaming encoding with arbitrary per-part headers and streaming decoding, so both are implemented in `multipart.py`.

**Streaming pipeline (no buffering):**
- Request: `_build_batch_parts` (generator) → `encode_multipart` (yields `bytes` lazily) → urllib3 chunked transfer encoding
- Response: `response.stream(65536)` → `iter_multipart_response` (incremental boundary detection across arbitrary chunks) → `_parse_batch_response` (generator) → caller receives `Iterator[ManyResponse]`

**Operation classification:**
- `bytes` Put bodies: compressed eagerly (required for size-based batching), then classified — ≤1 MB goes in a batch request, >1 MB goes individually via the standard PUT endpoint
- `IO[bytes]` Put bodies: always sent individually to avoid eager reading
- Get and Delete: always batched (zero body size)

**Concurrency:** `ThreadPoolExecutor` dispatches batch chunks and individual ops concurrently (default `concurrency=3`). Results arrive in completion order, matching the Rust client's `buffer_unordered` semantics. Sequential execution (`concurrency=1`) preserves input order.

Close FS-330

### DIFF
--- a/clients/python/docs/conf.py
+++ b/clients/python/docs/conf.py
@@ -25,3 +25,5 @@ autodoc_default_options = {
     "undoc-members": True,
     "show-inheritance": True,
 }
+
+suppress_warnings = ["ref.python"]
--- a/clients/python/docs/objectstore_client.rst
+++ b/clients/python/docs/objectstore_client.rst
@@ -49,6 +49,14 @@ objectstore\_client.metrics module
    :show-inheritance:
    :undoc-members:
 
+objectstore\_client.multipart module
+------------------------------------
+
+.. automodule:: objectstore_client.multipart
+   :members:
+   :show-inheritance:
+   :undoc-members:
+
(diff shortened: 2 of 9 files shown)


lcian:
bugbot run

lcian:
@sentry review

lcian:
bugbot run

jan-auer:
Closing in favor of #546

PR #423, opened 2026-04-08
feat(server): Accept auth token via query parameter

lcian:
For requests that go from the frontend (users' browser) to Objectstore, we currently have no way to authenticate properly.
This begins to solve that problem by introducing a `X-Objectstore-Auth` query parameter that, when present, carries auth information, and is checked as one of the possible places where the JWT is.
Our Django endpoints can then return URLs embedding such a query parameter, to provide the auth info to the client that will be able to fetch images/files/whatever from the proxy endpoint directly.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2547,6 +2547,7 @@ dependencies = [
  "bytes",
  "elegant-departure",
  "figment",
+ "form_urlencoded",
  "futures",
  "futures-util",
  "globset",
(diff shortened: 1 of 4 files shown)


jan-auer (commented):
I'm +1 on this, but if we can push for self-signed URLs right away, we should do that instead and not support auth in query params at all. Do you see a straight forward way to implement this?

lcian:
Let's try to implement proper signed URLs.

PR #424, opened 2026-04-09
feat: Add pre-signed URL support for GET/HEAD requests

lcian:
Implements pre-signed URLs that grant time-limited access to specific objects, as an alternative to providing a JWT header.
We will need this for frontend components that rely on Objectstore to provide files (currently only snapshot images and diffs) and we could also use it for e.g. attachments downloads.

Pre-signed URLs are of the form:
```
/v1/objects/{usecase}/{scopes}/{key}?X-Os-Expires={unix_ts}&X-Os-KeyId={kid}&X-Os-Signature={base64url_sig}
```

- The signature is a Ed25519 signature of the path portion of the URL and query parameters. The signature is computed on a canonicalized combination of path and query params, to ensure that signature computation and verification are consistent.
- Pre-signed URLs are only supported for GET and HEAD requests for now. The rationale for that is to solve the immediate problem of GET requests coming from browsers, and avoid expanding the scope to PUT/POST which would have a more complex implementation (implementing those correctly would require signing (part of) the request headers).

Clients expose a `presign_url` utility which takes a private key and creates pre-signed URLs with a customizable TTL, 5 minutes by default.

Introduces a new `objectstore-shared` crate to house the shared signing logic and constants between the client and server crate. We already have `objectstore-types` that's similarly shared, but these are not really types, so I think a separate crate is warranted.

### DIFF
--- a/objectstore-server/src/extractors/service.rs
+++ b/objectstore-server/src/extractors/service.rs
@@ -1,7 +1,8 @@
-use axum::extract::FromRequestParts;
-use axum::http::{header, request::Parts};
+use axum::extract::{FromRequestParts, OriginalUri};
+use axum::http::{Uri, header, request::Parts};
 
-use crate::auth::{AuthAwareService, AuthContext};
+use crate::auth::presigned::extract_presigned_params;
+use crate::auth::{AuthAwareService, AuthContext, AuthError};
 use crate::endpoints::common::ApiError;
 use crate::state::ServiceState;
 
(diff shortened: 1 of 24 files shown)


objectstore-server/src/extractors/service.rs
```diff
@@ -20,19 +21,35 @@ impl FromRequestParts<ServiceState> for AuthAwareService {
...
+                params,
+                &parts.method,
+                &parts.uri,
+                &state.key_directory,
+            ),
+            // Fall back to header-based JWT auth
+            (None, Some(jwt)) => AuthContext::from_encoded_jwt(jwt, &state.key_directory),
+            (None, None) => Err(AuthError::BadRequest("No authorization provided")),
+        };
+
```
lcian, line 48: We need the pre-signed auth to take precedence, because browsers will make requests to the proxy endpoint (or to Objectstore directly, if it were exposed through an NGINX rule) with both an `Authorization` header and the pre-signature.

Another approach would be to try all 3 methods (2 headers + pre-signature) every time. Or even build a composite `AuthContext` that takes into account all authentication methods, if multiple of them are present together.
That's more resilient but also more expensive, I would prefer to avoid it.

lcian:
bugbot run

lcian:
@sentry review

lcian:
bugbot run

lcian:
This is complex:
- Needs a separate mechanism (other than the JWT) to sign and verify
- The mechanism is sensitive to encoding, ordering of query params etc.
The only benefit that this provides is fixed-size URLs.
The more sensible approach I will now implement is to extend JWTs to support an optional ObjectKey to scope down authorization further.

PR #425, opened 2026-04-09
feat(options): Derive SentryOptions for typed, live-reloading runtime configuration

jan-auer:
Separates the generic machinery for typed sentry-options from the Objectstore-specific option types, and introduces a `#[derive(SentryOptions)]` macro so that adding a new options struct requires no boilerplate.

The work is split following the `serde`/`serde_derive` pattern into two new crates:

- **`objectstore-typed-options`** — the `SentryOptions` trait, shared `Error` type, and background refresh loop. Exposes a `derive` feature flag that re-exports the proc macro.
- **`objectstore-typed-options-derive`** — the proc macro, which generates the singleton, `get()`, `init()`, and test override support for any annotated struct.

`objectstore-options` is then reduced to the concrete type definitions, with most of its dependencies removed.

> **Note:** `objectstore-typed-options` is a prototype intended to be upstreamed into `sentry-options` itself. The crate boundary makes it easy to iterate on the API before that happens.


Cargo.toml
```diff
@@ -26,6 +26,8 @@ objectstore-log = { path = "objectstore-log" }
 objectstore-metrics = { path = "objectstore-metrics" }
 objectstore-options = { path = "objectstore-options" }
 objectstore-server = { path = "objectstore-server" }
+objectstore-typed-options = { path = "objectstore-typed-options" }
+objectstore-typed-options-derive = { path = "objectstore-typed-options-derive" }
 objectstore-service = { path = "objectstore-service" }
 objectstore-test = { path = "objectstore-test" }
 objectstore-types = { path = "objectstore-types", version = "0.1.12" }
```
lcian: Nit
```suggestion
objectstore-service = { path = "objectstore-service" }
objectstore-test = { path = "objectstore-test" }
objectstore-typed-options = { path = "objectstore-typed-options" }
objectstore-typed-options-derive = { path = "objectstore-typed-options-derive" }
objectstore-types = { path = "objectstore-types", version = "0.1.12" }
```

objectstore-typed-options-derive/src/lib.rs
```diff
@@ -0,0 +1,298 @@
...
+                }
+                ::std::result::Result::Err(ref err) => {
+                    ::objectstore_typed_options::log_error(
+                        #namespace_str, "failed to refresh options", err,
+                    )
+                }
+            }
+        }
+
+        impl ::objectstore_typed_options::SentryOptions for #name {
```
lcian, line 183:
```suggestion
        #[automatically_derived]
        impl ::objectstore_typed_options::SentryOptions for #name {
```

objectstore-typed-options-derive/src/lib.rs
```diff
@@ -0,0 +1,298 @@
...
+            else {
+                return;
+            };
+
+            match <#name as ::objectstore_typed_options::SentryOptions>::deserialize(inner) {
+                ::std::result::Result::Ok(new_snapshot) => {
+                    snapshot.store(::std::sync::Arc::new(new_snapshot))
+                }
+                ::std::result::Result::Err(ref err) => {
+                    ::objectstore_typed_options::log_error(
```
lcian, line 175: I don't understand what the vision is for upstreaming this, but seems it would be nice if the user could provide their own callback to log the errors as they see fit, maybe with an attribute like
```
#[sentry_options(namespace = "objectstore", path = "../../sentry-options", error_logger = "my_logger")]
```

objectstore-typed-options-derive/Cargo.toml
```diff
@@ -0,0 +1,18 @@
...
+edition = "2024"
+publish = false
+
+[lib]
+proc-macro = true
+
+[dependencies]
+proc-macro2 = "1.0.95"
+quote = "1.0.40"
+syn = { version = "2.0.101", features = ["full"] }
```
lcian:
```suggestion
syn = "2.0.101"
```

lcian (commented):
The PR doesn't achieve its stated goal of "live-reloading" options.
We register `__refresh` as the callback on the `sentry_options::Options`, but that callback actually only fires when `Options::get` is called (which itself "synchronously" (not on a bg task) reloads the values if needed, then calls the callback).
Our code never calls `Options::get`, which means that the callback never fires, and the options are never reloaded.

We need a background task that calls `__INNER.get()` periodically and swaps the value in the `ArcSwap` -- at that point the callback becomes effectively useless.
Or, `sentry-options` need to be changed to introduce this background task and change the semantics of the refresh and the callback.

objectstore-typed-options-derive/src/lib.rs
jan-auer: great suggestion, thank you!

right now, `sentry-options` doesn't use tracing or logging at all. In https://github.com/getsentry/sentry-options/pull/116, I'm replacing this all with `tracing`, at which point hopefully this error doesn't have to be customized anymore. 

It may still make sense to add more customization though, such as another hook for the propagation callback. Those would be added upstream, as we don't need them here.

objectstore-server/src/killswitches.rs
```diff
@@ -7,23 +7,58 @@
...
+            cache: ThreadLocal::new(),
+        }
+    }
+
     /// Returns `true` if any of the contained killswitches matches the given context.
     ///
     /// On match, emits a `server.request.killswitched` metric counter and a `warn!` log.
     pub fn matches(&self, context: &ObjectContext, service: Option<&str>) -> bool {
-        let Some(killswitch) = self.find(context, service) else {
+        let options = Options::get();
```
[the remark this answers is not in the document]
jan-auer: Production code runs `init` during startup, in tests we always have the `testing` feature enabled through dev-dependencies. It's in the nature of this crate that init must be called.

objectstore-typed-options-derive/src/lib.rs
```diff
@@ -0,0 +1,299 @@
...
+                // accessed.
+                let inner = ::objectstore_typed_options::sentry_options::Options::builder()
+                    .with_schemas(&[(Self::NAMESPACE, Self::SCHEMA)])
+                    .with_callback(__refresh)
+                    .build()?;
+
+                __OPTIONS
+                    .set(::objectstore_typed_options::arc_swap::ArcSwap::from_pointee(
+                        Self::deserialize(&inner)?,
+                    ))
```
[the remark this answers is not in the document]
jan-auer: False. Global is only set after deserialization works. Also, we only call init once.

jan-auer:
@lcian thanks, that's a big one! Since this is an existing issue with #421, I'm inclined to merge it as is and then follow up. Maybe we can re-introduce the background thread, which would be ideal.

lcian (approved):
Well this way the options will effectively never update and we'll never incur the cost of refreshing them, so I'm okay with merging this and bumping the library later when the background refresh is upstreamed.

PR #429, opened 2026-04-13
feat!(auth): Consider ordering of scopes for auth

lcian:
This changes the format for the `scopes` field in our JWTs to be a `Vec`, and `AuthContext` to take ordering into account when validating auth.
Previously, `scopes` was a `map`, which meant that order didn't matter.

I think this is the way this was intended to work in the first place.
From https://github.com/getsentry/objectstore/pull/199:
> - an AuthContext with a scope of org.123/proj.456 covers objects whose scope is org.123/proj.456
> - an AuthContext with a scope of just org.123 covers all objects whose scope *starts* with org.123 (i.e., objects from any project in that org)

This is a breaking change as the serialization format for our JWTs changes.
We will need to update our client libraries in all services.
Luckily we're still not enforcing auth, so we can easily make the breaking change now.

### DIFF
--- a/objectstore-server/src/auth/context.rs
+++ b/objectstore-server/src/auth/context.rs
@@ -406,7 +406,7 @@ MC4CAQAwBQYDK2VwBCIEIKwVoE4TmTfWoqH3HgLVsEcHs9PHNe+ar/Hp6e4To8pK
     //   auth_context: org.123 / proj.456
     //         object: org.123
     #[test]
-    fn test_assert_authorized_shorter_request_scope_fails() -> Result<(), AuthError> {
+    fn test_assert_authorized_less_specific_request_scope_fails() -> Result<(), AuthError> {
         let auth_context = sample_auth_context("123", "456", max_permission());
         let object = ObjectContext {
             usecase: "attachments".into(),
@@ -419,8 +419,11 @@ MC4CAQAwBQYDK2VwBCIEIKwVoE4TmTfWoqH3HgLVsEcHs9PHNe+ar/Hp6e4To8pK
         Ok(())
     }
 
+    // Allowed:
+    //   auth_context: org.123 / proj.*
+    //         object: org.123 / proj.456 / extra.789
     #[test]
-    fn test_assert_authorized_extra_request_scope_allowed() -> Result<(), AuthError> {
+    fn test_assert_authorized_more_specific_request_scope_succeeds() -> Result<(), AuthError> {
         let auth_context = sample_auth_context("123", "*", max_permission());
         let object = ObjectContext {
             usecase: "attachments".into(),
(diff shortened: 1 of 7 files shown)


objectstore-server/src/auth/context.rs
```diff
@@ -148,12 +161,18 @@ impl AuthContext {
...
+        }
+
+        for ((authorized_name, authorized_value), request_scope) in
+            self.scopes.iter().zip(&context.scopes)
+        {
+            let authorized = authorized_name == request_scope.name()
+                && match authorized_value {
+                    StringOrWildcard::String(s) => s == request_scope.value(),
+                    StringOrWildcard::Wildcard => true,
+                };
```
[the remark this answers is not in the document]
lcian: This is intended and I added a test for it. Definitely something we need to discuss though.

objectstore-server/src/auth/context.rs
```diff
@@ -373,22 +399,83 @@ MC4CAQAwBQYDK2VwBCIEIKwVoE4TmTfWoqH3HgLVsEcHs9PHNe+ar/Hp6e4To8pK
...
         auth_context.assert_authorized(Permission::ObjectRead, &object)?;
 
         Ok(())
     }
 
+    // Allowed:
+    //   auth_context: (empty vec)
+    //         object: org.123
+    #[test]
+    fn test_assert_authorized_empty_scope_allows_any_request() -> Result<(), AuthError> {
```
lcian, line 449: This is weird, but consistent with what we want semantically.
It could be used to grant access to the whole usecase.

matt-codecov: currently you can do this with a token for `org.*/proj.*`

matt-codecov (commented):
iirc i actually made scopes unordered on purpose in an attempt to preserve the service-agnostic design of the token. it is less awkward for other services to ignore objectstore-specific scopes if the token doesn't encode those objectstore scopes as the root of a hierarchy

scope order/hierarchy is now up to the application to interpret. for objectstore, the interpretation is:
1. the `os:usecase` scope is required and its value must match the first part of an object's scope
2. every subsequent part of the object's scope must be present in the token's scopes
3. any token scope that isn't present in the object's scope is ignored
4. wildcards must be explicit; `org.123` does not imply `org.123/proj.*`

i guess i never updated the PR description or notion doc (whoops! sorry) but the tests show the expected behavior: https://github.com/getsentry/objectstore/blob/af57fd8f1a1377129a77eac3bb2e86011fc9dfc4/objectstore-server/src/auth/context.rs#L350-L413

#\3 above is a bit awkward. consider this possibly counterintuitive example:
```rust
// Allowed:
//   auth_context: org.123 / proj.456
//         object: org.123
```
because there is no project component in the object scope, any project component in the token scope is ignored. it behaves this way for two reasons:
- this is how objectstore can ignore scopes specific to other services
- something like debug symbols may be uploaded at org scope yet still be accessible to multiple projects that need to read them to symbolicate errors

that said, i admit it is weird.

all this said: do you still want to move forward with this change? i'm not necessarily opposed. sentry does need a real service-agnostic auth scheme, but if i pick that effort up again it will be on another team.

objectstore-server/src/auth/context.rs
lcian: In theory we don't enforce the scope to be of the form `org.*/proj.*`.

lcian:
Not prioritizing this for now, we can always change this with relatively the same friction.

PR #444, opened 2026-04-27
chore(deps): Bump bigtable_rs to latest rev

lcian:
Includes https://github.com/getsentry/bigtable_rs/pull/8.
I noticed sometimes we were getting keepalive timeouts in prod, so maybe this improves things as we will ping more often.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -430,7 +430,7 @@ dependencies = [
 [[package]]
 name = "bigtable_rs"
 version = "0.2.19"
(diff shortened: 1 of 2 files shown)


matt-codecov (approved):
can we nudge upstream to release so we can get back on the crates.io dep?

lcian:
> can we nudge upstream to release so we can get back on the crates.io dep?

We can do so but that's not enough, in our fork we also have the pre-emptive refresh of channels which I haven't PRd upstream yet, I've kinda been procrastinating it as it will most likely need some discussion around API design and implementation.

PR #445, opened 2026-04-27
feat(service): Add MultipartUploadBackend trait and related types

lcian:
⚠️ Stacked on https://github.com/getsentry/objectstore/pull/442

Adds a trait defining `MultipartUploadBackend` and its associated types according to the [design doc](https://www.notion.so/sentry/Design-Multipart-uploads-3438b10e4b5d8009aeedcd8db91122ea?source=copy_link).

The next PR will implement `MultipartUploadBackend` on the `s3_compatible` backend.
After that and implementations for `GCS` and `TieredStorage` land, we will change `TieredStorage` to require a `MultipartUploadBackend` in its `long_term` slot, as opposed to just a `Backend`.

### DIFF
--- a/objectstore-service/src/multipart.rs
+++ b/objectstore-service/src/multipart.rs
@@ -1,0 +1,66 @@
+//! Shared types for Objectstore's multipart upload protocol.
+
+use std::time::SystemTime;
+
+/// Identifier for an in-progress multipart upload.
+pub type UploadId = String;
+/// 1-indexed position of a part within its multipart upload.
+pub type PartNumber = u32;
+/// Opaque per-part identifier returned by the backend after a successful part upload.
+pub type ETag = String;
+
+/// Description of one part in the response to
+/// [`MultipartUploadBackend::list_parts`](crate::backend::common::MultipartUploadBackend::list_parts).
+#[derive(Clone, Debug)]
+pub struct PartInfo {
+    /// 1-indexed position of this part within the upload.
+    pub part_number: PartNumber,
+    /// Identifier returned when the part was uploaded.
+    pub etag: ETag,
+    /// Server-recorded time at which the part was uploaded.
+    pub last_modified: SystemTime,
+    /// Size of the part in bytes.
+    pub size: u64,
+}
+
+/// Pair of (part number, ETag) that the client provides on
+/// [`MultipartUploadBackend::complete_multipart`](crate::backend::common::MultipartUploadBackend::complete_multipart)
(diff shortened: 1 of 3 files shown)


objectstore-service/src/multipart.rs
```diff
@@ -0,0 +1,66 @@
...
+/// Description of one part in the response to
+/// [`MultipartUploadBackend::list_parts`](crate::backend::common::MultipartUploadBackend::list_parts).
+#[derive(Clone, Debug)]
+pub struct PartInfo {
+    /// 1-indexed position of this part within the upload.
+    pub part_number: PartNumber,
+    /// Identifier returned when the part was uploaded.
+    pub etag: ETag,
+    /// Server-recorded time at which the part was uploaded.
+    pub last_modified: SystemTime,
```
matt-codecov, line 21: if this is communicated back to users should it be `DateTime<Utc>` or something?

objectstore-service/src/multipart.rs
```diff
@@ -0,0 +1,66 @@
...
+
+/// Backend response for
+/// [`MultipartUploadBackend::initiate_multipart`](crate::backend::common::MultipartUploadBackend::initiate_multipart).
+pub type InitiateMultipartResponse = UploadId;
+/// Backend response for
+/// [`MultipartUploadBackend::upload_part`](crate::backend::common::MultipartUploadBackend::upload_part).
+pub type UploadPartResponse = ETag;
+/// Backend response for
+/// [`MultipartUploadBackend::list_parts`](crate::backend::common::MultipartUploadBackend::list_parts).
+pub type ListPartsResponse = ListedParts;
```
matt-codecov, line 60: am i missing an `IntoResponse` impl or something for this?

matt-codecov (approved):

objectstore-service/src/multipart.rs
lcian: Not sure if we need it or that's automatic, we probably just need it to be `Serialize`? 
Will take a look at it when I implement the objectstore-server part (endpoints).

objectstore-service/src/multipart.rs
lcian: Good catch!
We can just add `#[serde(with = "humantime_serde")]` once we `#[derive(Serialize)]` on this, that returns an RFC 3339 string and is consistent with how we surface timestamps when serializing `Metadata` to headers.
I will add `Serialize` to this when I implement the endpoints.

PR #446, opened 2026-04-29
feat(s3): Add auth configuration

lcian:
⚠️ Stacked on https://github.com/getsentry/objectstore/pull/445

We don't really need this anymore because I found out my GCS -> S3 approach won't really work.
However this will be needed for self-hosted and I already had this done, so might as well merge it.


lcian:
Ready for review but not priority anymore.

lcian:
Let's close this for now. We might or might not be able to reuse this when we pick up S3 support again.

PR #447, opened 2026-04-29
feat(auth): Default auth enforcement to true

lcian:
This flips the default from opt-in to opt-out so config or templating issues fail closed instead of fail open. 
The explicit `enforce` option is kept so local dev, sandbox, and devservices can disable auth when needed.
The server now fails at startup if `enforce: true` and no keys are configured.

Close FS-347

### DIFF
--- a/clients/rust/tests/e2e.rs
+++ b/clients/rust/tests/e2e.rs
@@ -172,16 +172,12 @@ async fn stores_under_given_key() {
 
 #[tokio::test]
 async fn stores_structured_keys() {
-    let server = TestServer::with_config(config::Config {
-        auth: config::AuthZ {
-            enforce: false,
-            ..Default::default()
-        },
-        ..Default::default()
-    })
-    .await;
+    let server = test_server().await;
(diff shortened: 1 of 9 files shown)


clients/rust/tests/e2e.rs
```diff
@@ -172,7 +172,14 @@ async fn stores_under_given_key() {
...
+    let server = TestServer::with_config(config::Config {
+        auth: config::AuthZ {
+            enforce: false,
+            ..Default::default()
+        },
+        ..Default::default()
+    })
+    .await;
 
     let client = Client::builder(server.url("/")).build().unwrap();
```
matt-codecov: is there a reason we don't add `.token(test_token_generator()` here and then just use the `test_server()` helper to start a test server?

matt-codecov (approved):

clients/rust/tests/e2e.rs
lcian: Fixed

PR #449, opened 2026-04-29
ci: Report Test Coverage

lcian:
Adds setup for test coverage using Codecov.
Coverage is collected using the complete test suite (including emulators), so that we can measure coverage of the backends too.
Status checks are reported as informational only, they don't block PRs from merging.

### DIFF
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -1,0 +1,10 @@
+coverage:
+  status:
+    project:
+      default:
+        informational: true
+    patch: false
+
+comment:
+  layout: "diff"
+  require_changes: true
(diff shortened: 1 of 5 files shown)


matt-codecov (approved):
may as well add python client test coverage too

https://github.com/codecov/codecov-rs/blob/main/codecov.yml is an example file that defines "components" so you can track coverage separately for different parts of your project if we want to do that to separate our clients and backend

lcian:
> may as well add python client test coverage too
> 
> [codecov/codecov-rs@`main`/codecov.yml](https://github.com/codecov/codecov-rs/blob/main/codecov.yml) is an example file that defines "components" so you can track coverage separately for different parts of your project if we want to do that to separate our clients and backend

Nice, added it!

PR #450, opened 2026-04-29
feat(service): Implement MultipartUploadBackend on GcsBackend

lcian:
Implements multipart uploads on the GcsBackend, using the S3-compatible XML API.
Ideally this implementation could go in `s3_compatible.rs` and be shared, which I've attempted, but this wasn't that easy due to auth differences between the two.

Also reduces some indirection in the service types.

A follow-up PR will add tests against a fork of our emulator (Google's `storage-testbench`) where I implemented this API.

Closes FS-338

### DIFF
--- a/objectstore-service/src/backend/gcs.rs
+++ b/objectstore-service/src/backend/gcs.rs
@@ -293,59 +293,6 @@ impl serde::Serialize for GcsMetaKey {
     }
 }
 
-/// Response for a GCS XML API initiate multipart request.
-#[derive(Debug, Deserialize)]
-#[serde(rename_all = "PascalCase")]
-struct XmlInitiateMultipartUploadResponse {
-    upload_id: String,
-}
-
-/// Response for a GCS XML API list parts request.
-#[derive(Debug, Deserialize)]
-#[serde(rename_all = "PascalCase")]
-struct XmlListPartsResponse {
-    #[serde(default)]
-    is_truncated: bool,
-    next_part_number_marker: Option<u32>,
-    #[serde(default, rename = "Part")]
-    parts: Vec<XmlPart>,
-}
-
-#[derive(Debug, Deserialize)]
-#[serde(rename_all = "PascalCase")]
-struct XmlPart {
-    part_number: u32,
-    #[serde(rename = "ETag")]
-    e_tag: String,
-    #[serde(with = "humantime_serde")]
-    last_modified: SystemTime,
-    size: u64,
-}
-
-/// Request body for a GCS XML API complete multipart upload request.
-#[derive(Debug, Serialize)]
-#[serde(rename = "CompleteMultipartUpload")]
-struct XmlCompleteMultipartUpload<'a> {
-    #[serde(rename = "Part")]
-    parts: Vec<XmlCompletePart<'a>>,
-}
-
-#[derive(Debug, Serialize)]
-#[serde(rename_all = "PascalCase")]
-struct XmlCompletePart<'a> {
-    part_number: PartNumber,
-    #[serde(rename = "ETag")]
-    e_tag: &'a str,
-}
-
-/// Error response for a complete multipart request.
-#[derive(Debug, Deserialize)]
-#[serde(rename = "Error", rename_all = "PascalCase")]
-struct XmlError {
-    code: String,
-    message: String,
-}
-
 /// Builds HTTP headers that encode `metadata` for a GCS XML API request.
 fn metadata_to_gcs_headers(metadata: &Metadata) -> Result<header::HeaderMap> {
     let mut headers = header::HeaderMap::new();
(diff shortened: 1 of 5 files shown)


objectstore-service/src/backend/gcs.rs
```diff
@@ -286,6 +292,126 @@ impl serde::Serialize for GcsMetaKey {
...
+
+    if let Some(origin) = &metadata.origin {
+        insert_gcs_meta_header(&mut headers, &GcsMetaKey::Origin, origin)?;
+    }
+
+    for (key, value) in &metadata.custom {
+        insert_gcs_meta_header(&mut headers, &GcsMetaKey::Custom(key.clone()), value)?;
+    }
+
+    Ok(headers)
```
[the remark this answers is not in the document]
lcian: Yes, as you said one serializes to JSON and the other to headers, so we cannot really share the logic.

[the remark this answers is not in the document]
matt-codecov: this logic is basically duplicated in a third place but for a slightly different purpose:
https://github.com/getsentry/objectstore/blob/38ab344e326d3352484a762d6c991a3ff1125de0/objectstore-types/src/metadata.rs#L339

it would be nice to do something about this, but i don't have a clear easy idea. we could maybe `impl Serialize<...> for Metadata` in a map style manually and do the humantime and conditional stuff in there and then write `HeaderMetaDeserializer`, `XmlHeaderMetaDeserializer`, `JsonMetaDeserializer` to deal with the field renames

objectstore-service/src/backend/gcs.rs
```diff
@@ -377,6 +503,30 @@ impl GcsBackend {
...
+                    self.endpoint
+                ),
+                cause: None,
+            })?;
+            segments.push(&self.bucket);
+            for part in id.as_storage_path().to_string().split('/') {
+                segments.push(part);
+            }
+        }
+        Ok(url)
```
matt-codecov, line 474: no real preference, but you could one-shot this by concatenating and re-parsing
```rust
Ok(Url::parse(&format!(
    "{}/{}/{}",
    self.endpoint.as_str(),
    &self.bucket,
    id.as_storage_path(),
))?)
```

objectstore-service/src/backend/gcs.rs
```diff
@@ -286,6 +292,126 @@ impl serde::Serialize for GcsMetaKey {
...
+#[derive(Debug, Deserialize)]
+#[serde(rename_all = "PascalCase")]
+struct XmlPart {
+    part_number: u32,
+    #[serde(rename = "ETag")]
+    e_tag: String,
+    #[serde(with = "humantime_serde")]
+    last_modified: SystemTime,
+    size: u64,
+}
```
matt-codecov: is PascalCase the only reason we need duplicate types for these? that's a bummer

maybe `impl From<Xml*> for *` for these types to keep the handlers tidy
```rust
impl From<XmlPart> for Part {
    fn from(part: XmlPart) -> Self {
        Self {
            ...
        }
    }
}

impl From<XmlListPartsResponse> for ListPartsResponse {
    fn from(xml: XmlListPartsResponse) -> Self {
        Self {
            ...
            parts: xml.parts.into_iter().map(|part| part.into()).collect(),
        }
    }
}
```

objectstore-service/src/backend/gcs.rs
```diff
@@ -621,6 +771,220 @@ impl Backend for GcsBackend {
...
+                last_modified: p.last_modified,
+                size: p.size,
+            })
+            .collect();
+
+        Ok(ListPartsResponse {
+            parts,
+            is_truncated: parsed.is_truncated,
+            next_part_number_marker: parsed.next_part_number_marker,
+        })
```
matt-codecov: with the `impl From<Xml*> for *` suggestion, all this could be replaced with
```rust
let xml_resp = quick_xml::de::from_reader(body.as_ref()).map_err(|e| Error::Generic {
    context: "GCS: failed to parse list parts response".to_owned(),
    cause: Some(Box::new(e)),
})?;
Ok(xml_resp.into())
```

matt-codecov (approved):
accepted with optional comments

objectstore-service/src/backend/gcs.rs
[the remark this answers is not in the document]
lcian: Great idea, I didn't think of that, we can definitely use some custom `Serialize` implementations or roll our own trait based on visitors, filed a follow up ticket: https://linear.app/getsentry/issue/FS-353/unify-metadata-serialization-logic

objectstore-service/src/backend/gcs.rs
lcian: That's better, yes.

objectstore-service/src/backend/gcs.rs
lcian: Good suggestion, addressed.

PR #454, opened 2026-04-30
test: Add tests for GCS multipart uploads

lcian:
⚠️ Stacked on https://github.com/getsentry/objectstore/pull/450

Adds e2e tests against our fork of https://github.com/googleapis/storage-testbench, which contains an XML API multipart upload implementation https://github.com/getsentry/google-cloud-storage-testbench
Also switches devservices to use that fork.

The implementation present in this fork has been completely written by LLMs, so it should not be trusted for anything but catching obvious regressions if we ever make changes to the GCS backend.
Any non-trivial change to the GCS backend that touches the multipart upload implementation should be tested in sandbox against real GCS before merging.
I've added a code comment about this too.

Part of https://github.com/getsentry/objectstore/pull/450, will be merged into it.
Part of FS-338

PR #456, opened 2026-05-04
feat(service): Implement MultipartUploadBackend on InMemory and LocalFs

lcian:
⚠️ Stacked on https://github.com/getsentry/objectstore/pull/454

Implements `MultipartUploadBackend` on `InMemoryBackend` and `LocalFsBackend`, used for tests.
`TieredStorage` will come in the next PR.

`InMemoryBackend` now stores some auxiliary data structures dedicated to multipart uploads, while `LocalFs` uses a `__multipart__/` directory where to store metadata and parts for each ongoing upload.

Close FS-349

### DIFF
--- a/objectstore-service/src/backend/in_memory.rs
+++ b/objectstore-service/src/backend/in_memory.rs
@@ -329,47 +329,59 @@ impl MultipartUploadBackend for InMemoryBackend {
         upload_id: &UploadId,
         parts: Vec<CompletedPart>,
     ) -> Result<CompleteMultipartResponse> {
-        let upload = self
-            .multipart_store
-            .lock()
-            .unwrap()
-            .remove(&(id.clone(), upload_id.clone()))
-            .ok_or_else(|| Error::generic("multipart upload not found"))?;
+        let key = (id.clone(), upload_id.clone());
 
-        for completed in &parts {
-            match upload.parts.get(&completed.part_number) {
-                None => {
-                    return Ok(Some(crate::multipart::CompleteMultipartError {
-                        code: "InvalidPart".into(),
-                        message: format!("part number {} was not uploaded", completed.part_number),
-                    }));
+        // Validate and assemble while holding the multipart lock, but don't
+        // remove the upload yet — a failed validation must leave it intact so
+        // the client can retry.
+        let assembled = {
+            let store = self.multipart_store.lock().unwrap();
+            let upload = store
+                .get(&key)
+                .ok_or_else(|| Error::generic("multipart upload not found"))?;
+
+            for completed in &parts {
+                match upload.parts.get(&completed.part_number) {
+                    None => {
+                        return Ok(Some(crate::multipart::CompleteMultipartError {
+                            code: "InvalidPart".into(),
+                            message: format!(
+                                "part number {} was not uploaded",
+                                completed.part_number
+                            ),
+                        }));
+                    }
+                    Some(stored) if stored.etag != completed.etag => {
+                        return Ok(Some(crate::multipart::CompleteMultipartError {
+                            code: "InvalidPart".into(),
+                            message: format!(
+                                "etag mismatch for part {}: expected {}, got {}",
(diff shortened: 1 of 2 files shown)


objectstore-service/src/backend/local_fs.rs
```diff
@@ -152,6 +160,226 @@ impl Backend for LocalFsBackend {
...
+        }
+
+        parts.sort_by_key(|p| p.part_number);
+
+        let max = max_parts.unwrap_or(u32::MAX) as usize;
+        let is_truncated = parts.len() > max;
+        parts.truncate(max);
+
+        let next_part_number_marker = if is_truncated {
+            parts.last().map(|p| p.part_number)
```
matt-codecov, line 305: so the marker isn't the first part number to include, it's the last part number to skip? both impls agree on that, and i can imagine why it is this way, just confirming

objectstore-service/src/backend/in_memory.rs
```diff
@@ -195,6 +217,176 @@ impl HighVolumeBackend for InMemoryBackend {
...
+        id: &ObjectId,
+        metadata: &Metadata,
+    ) -> Result<InitiateMultipartResponse> {
+        let upload_id = uuid::Uuid::now_v7().to_string();
+        let upload = MultipartUpload {
+            metadata: metadata.clone(),
+            parts: BTreeMap::new(),
+        };
+        self.multipart_store
+            .lock()
```
matt-codecov, line 233: every operation taking a global lock on the store is probably not suitable for production. if we need this to be production-worthy we should look into in-memory duckdb or something

objectstore-service/src/backend/local_fs.rs
```diff
@@ -152,6 +160,226 @@ impl Backend for LocalFsBackend {
...
+    fn multipart_dir(&self, id: &ObjectId, upload_id: &UploadId) -> PathBuf {
+        self.path
+            .join("__multipart__")
+            .join(id.as_storage_path().to_string())
+            .join(upload_id)
+    }
+}
+
+#[async_trait::async_trait]
+impl MultipartUploadBackend for LocalFsBackend {
```
matt-codecov, line 172: i think this impl will be pretty susceptible to 5XXs. operations don't acquire any sort of lock on the upload dir so a concurrent `abort`/`complete` request can remove the upload dir in the middle of other concurrent handlers. if/when we want this backend to be production-worthy we probably want to make sure races result in an appropriate 4XX

matt-codecov (approved):

objectstore-service/src/backend/in_memory.rs
lcian: This backend is only used in tests, it's not intended to be used in prod (local_fs kinda is, on the other hand).

objectstore-service/src/backend/local_fs.rs
lcian: Hmm yes I see what you mean. We need some sort of lock files. Filed as a follow-up we need to eventually do for self-hosted support: https://linear.app/getsentry/issue/PF-81/avoid-races-in-fs-backend

objectstore-service/src/backend/local_fs.rs
lcian: Yep!

objectstore-service/src/backend/local_fs.rs
```diff
@@ -152,6 +159,267 @@ impl Backend for LocalFsBackend {
...
+        content_length: u64,
+        _content_md5: Option<&str>,
+        body: ClientStream,
+    ) -> Result<UploadPartResponse> {
+        let dir = self.multipart_dir(id, upload_id);
+        if !tokio::fs::try_exists(&dir).await? {
+            return Err(Error::generic("multipart upload not found"));
+        }
+
+        let etag = format!("\"etag-{part_number}-{content_length}\"");
```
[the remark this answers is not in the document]
lcian: Etag is an opaque identifier, it doesn't have to be the same across backends.

PR #457, opened 2026-05-04
feat(rust-client): Make concurrency configurable in ManyBuilder

lcian:
Add `max_individual_concurrency` and `max_batch_concurrency` builder methods to `ManyBuilder`, allowing callers to tune how many HTTP requests are in-flight simultaneously.

We can potentially use this in combination with https://github.com/getsentry/sentry/blob/7256de9f017ce624db7de3bd760f2a5dcd244cf3/src/sentry/objectstore/types.py to tune the concurrency to use for snapshot uploads in all `sentry-cli`s, as a "remote config".

Part of FS-352

### DIFF
--- a/clients/rust/src/many.rs
+++ b/clients/rust/src/many.rs
@@ -30,13 +30,15 @@ const MAX_BATCH_OPS: usize = 1000;
 /// Maximum amount of bytes to send as a part's body in a batch request.
 const MAX_BATCH_PART_SIZE: u32 = 1024 * 1024; // 1 MB
 
-/// Operations that are guaranteed to exceed `MAX_BATCH_PART_SIZE` are executed using requests to
-/// the individual object endpoint, rather than the batch endpoint.
-/// This determines the maximum number of such requests that can be executed concurrently.
-const MAX_INDIVIDUAL_CONCURRENCY: usize = 5;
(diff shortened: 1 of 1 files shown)


clients/rust/src/many.rs
```diff
@@ -649,6 +655,12 @@ impl ManyBuilder {
     /// The results are not guaranteed to be in the order they were originally enqueued in.
     pub async fn send(self) -> OperationResults {
         let session = self.session;
+        let individual_concurrency = self
+            .max_individual_concurrency
+            .unwrap_or(DEFAULT_INDIVIDUAL_CONCURRENCY);
```
lcian:
```suggestion
            .unwrap_or(DEFAULT_INDIVIDUAL_CONCURRENCY)
            .max(1);
```

clients/rust/src/many.rs
```diff
@@ -649,6 +655,12 @@ impl ManyBuilder {
...
+        let batch_concurrency = self
+            .max_batch_concurrency
+            .unwrap_or(DEFAULT_BATCH_CONCURRENCY);
```
lcian:
```suggestion
            .unwrap_or(DEFAULT_BATCH_CONCURRENCY)
            .max(1);
```

matt-codecov (approved):

PR #458, opened 2026-05-04
feat: Implement MultipartUploadBackend on TieredStorage

lcian:
Last implementation of `MultipartUploadBackend` in `objectstore-service`.
Now `TieredStorage` requires a `MultipartUploadBackend` in its `long_term` slot.
All methods just pretty much forward to `long_term`, with the exception of `complete_multipart` which needs to use the `Changelog` for consistency and deal with the tombstones, with logic very similar to `put_object` except that we need an additional metadata read call.

Close FS-339


lcian:
This is still WIP.

objectstore-service/src/backend/tiered.rs
```diff
@@ -540,6 +547,192 @@ where
...
+impl TryFrom<&UploadId> for TieredUploadId {
+    type Error = Error;
+
+    fn try_from(value: &UploadId) -> Result<Self, Self::Error> {
+        let json = base64::engine::general_purpose::URL_SAFE_NO_PAD
+            .decode(value.as_bytes())
+            .map_err(|e| Error::generic(format!("invalid multipart upload ID: {e}")))?;
+        serde_json::from_slice(&json).map_err(|e| Error::serde("decoding multipart token", e))
+    }
+}
```
lcian, line 592: I've thought about whether we want to encrypt or otherwise obfuscate these contents to hide the physical revision and upload id from the user, but I don't think there's really any issue (security-wise or otherwise) with just giving them to the user in clear.

objectstore-service/src/backend/tiered.rs
```diff
@@ -540,6 +547,192 @@ where
...
+    upload_id: String,
+}
+
+impl TryInto<UploadId> for TieredUploadId {
+    type Error = Error;
+
+    fn try_into(self) -> Result<UploadId, Self::Error> {
+        use base64::Engine;
+        let json =
+            serde_json::to_vec(&self).map_err(|e| Error::serde("encoding multipart token", e))?;
```
lcian, line 578: Maybe there's a smarter way to concatenate the 2 with less size overhead than JSON...
But any separator could theoretically also appear in the ID returned by the inner backend.

lcian:
bugbot run

objectstore-service/src/backend/tiered.rs
```diff
@@ -540,6 +547,192 @@ where
...
+            })
+            .await?;
+
+        let error = self
+            .inner
+            .long_term
+            .complete_multipart(&physical, &tiered.upload_id, parts)
+            .await?;
+
+        if error.is_some() {
```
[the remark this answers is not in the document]
lcian: This is fine. The user needs to have the possibility to retry the `complete_multipart` call if they supplied wrong data.
Cleanup of stale ongoing multipart uploads will be handled by each long-term backend.

objectstore-service/src/backend/tiered.rs
```diff
@@ -540,6 +547,192 @@ where
...
+        // 3. Retrieve the metadata of the object, which was determined at initiation time, to
+        //    get the expiration policy.
+        let metadata = self
+            .inner
+            .long_term
+            .get_metadata(&physical)
+            .await?
+            .ok_or_else(|| {
+                Error::generic("completed multipart object not found in long-term storage")
+            })?;
```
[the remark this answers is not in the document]
lcian: It's very unlikely that this fails. I don't want to encode more information in the token.

objectstore-service/src/backend/tiered.rs
```diff
@@ -540,6 +547,192 @@ where
...
+struct TieredUploadId {
+    revision: String,
+    upload_id: String,
+}
+
+impl TryInto<UploadId> for TieredUploadId {
+    type Error = Error;
+
+    fn try_into(self) -> Result<UploadId, Self::Error> {
+        use base64::Engine;
```
lcian: This is weird, I'll move it up

objectstore-service/src/backend/changelog.rs
```diff
@@ -27,7 +27,7 @@ use std::time::Duration;
 use tokio_util::task::TaskTracker;
 use tokio_util::task::task_tracker::TaskTrackerToken;
 
-use crate::backend::common::{Backend, HighVolumeBackend, TieredMetadata};
+use crate::backend::common::{HighVolumeBackend, MultipartUploadBackend, TieredMetadata};
```
matt-codecov, line 30: is every long-term storage backend required to impl `MultipartUploadBackend`? should we rename it `LongTermBackend` trait to parallel the `HighVolumeBackend` and make sure both have a trait bound requiring that `Backend` be implemented too?

objectstore-service/src/backend/mod.rs
```diff
@@ -104,3 +104,28 @@ async fn hv_from_config(
...
+    FileSystem(local_fs::FileSystemConfig),
+
+    /// [Google Cloud Storage] backend (type `"gcs"`).
+    ///
+    /// [Google Cloud Storage]: https://cloud.google.com/storage
+    Gcs(gcs::GcsConfig),
+}
+
+/// Constructs a type-erased [`common::MultipartUploadBackend`] from the given config.
+async fn lt_from_config(
```
matt-codecov, line 124: i assume `lt` means `long_term` but it's a little awkward because none of the types around refer to "long term" storage. could you either expand the name or take the above renaming suggestion?

objectstore-service/src/backend/tiered.rs
```diff
@@ -540,6 +547,192 @@ where
...
+            .await?;
+
+        let error = self
+            .inner
+            .long_term
+            .complete_multipart(&physical, &tiered.upload_id, parts)
+            .await?;
+
+        if error.is_some() {
+            return Ok(error);
```
matt-codecov: is `Ok` appropriate here? is this an endpoint that returns 200 OK always and you have to read the response body to interpret errors?

objectstore-service/src/backend/tiered.rs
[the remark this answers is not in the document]
matt-codecov: another option is to move the `inner.long_term.get_metadata()` call earlier in the function before we create our change guard. that way if it fails we can at least 5XX and the user can retry before we've made any persistent changes

looking at the change guard impl
https://github.com/getsentry/objectstore/blob/1615f71d97624af34f73a749555193d4884936eb/objectstore-service/src/backend/changelog.rs#L304-L330

thinking through the remaining possible errors and how the change guard would handle them:
- if `inner.long_term.complete_multipart()` returns an error:
  - the guard will drop in the `Recorded` without cleaning anything up. the user can retry, unless the underlying backend is in an unretryable state
- if `inner.high_volume.compare_and_write()` returns an error and didn't persist anything:
  - the guard will drop in the `Written` state. it will fetch the current tombstone ID, see that our new ID doesn't match, and delete the new ID from LT storage.
  - the user has to redo the whole multipart upload, but the system should be sane. we need to make sure the error communicates this
  - this is the same issue as what the bot describes, but this instance is unavoidable
- if `inner.high_volume.compare_and_write()` returns an error but persisted the tombstone:
  -  the guard will drop in the `Written` state. it will fetch the current tombstone ID, see that our new ID _does_ match, and exit without cleaning anything up.

incidentally i see our change log doesn't yet have a durable backend implemented. i guess that's what FS-310 is still open for

matt-codecov (approved):
stamp w comments

objectstore-service/src/backend/changelog.rs
lcian: Yes exactly, it's implemented as follows: https://github.com/getsentry/objectstore/blob/1615f71d97624af34f73a749555193d4884936eb/objectstore-service/src/backend/common.rs#L69
So, it already has a trait bound on `Backend`.

I guess a more abstracted name for it could be `LongTermBackend`, yeah. Let me change it in a follow-up.

objectstore-service/src/backend/mod.rs
lcian: Yep!

objectstore-service/src/backend/tiered.rs
lcian: Yes that's intended, that's also how S3/GCS implement this and I mentioned it in the design doc.
That's because the `complete` call can take a while (I guess it's assembling the object in the backend), so S3 takes exactly this approach of a 200 and error in the body, therefore the most general solution is for us to mirror it.
Otherwise, the client might very easily run into a timeout if we don't return a response/headers within X s.

lcian: But probably this warrants a comment explaining how it works.

objectstore-service/src/backend/tiered.rs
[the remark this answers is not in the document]
lcian: > incidentally i see our change log doesn't yet have a durable backend implemented. i guess that's what FS-310 is still open for

Exactly.

> another option is to move the inner.long_term.get_metadata() call earlier in the function before we create our change guard. that way if it fails we can at least 5XX and the user can retry before we've made any persistent changes

We cannot call `get_metadata` yet because the object doesn't exist. S3 and GCS don't have APIs to retrieve the metadata for an incomplete multipart upload.
That's why we need to materialize the object first. And therefore, we need to create the guard beforehand, to avoid leaving that blob orphaned.

The alternative would be to include the expiration policy in the `UploadId` we give back to the user, but that introduces all kinds of conditions you have to handle, what if the user passes a different expiration policy at creation/completion? Also, we would need to validate the expiration policy in two places, etc. etc. -- in my opinion the first call should be the source of truth for this.

But you're right that this warrants some more thought.
With the current implementation, the client can basically never retry `complete_multipart` effectivly, because the dropped `ChangeGuard` will probably delete the new blob even if it completed successfully 🤔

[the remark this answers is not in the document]
lcian: The core issue is that `ChangeGuard::Recorded` assumes that every LT write uses a unique key, while in this case we would retry multiple times on the same key (revision).

lcian:
This is pending resolution of https://github.com/getsentry/objectstore/pull/458#discussion_r3200217052

objectstore-service/src/backend/changelog.rs
lcian: Filed FS-373

objectstore-service/src/backend/tiered.rs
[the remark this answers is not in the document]
lcian: Opened https://github.com/getsentry/objectstore/pull/480 to address this.

lcian:
Turning this into a draft to avoid getting confused until https://github.com/getsentry/objectstore/pull/480 has been reviewed.

objectstore-service/src/backend/tiered.rs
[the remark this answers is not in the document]
lcian: #480 merged into this one (sorry but it was easier for the stacking)

PR #459, opened 2026-05-05
fix(py-client): Unpin `uv_build`

lcian:
Reverts https://github.com/getsentry/objectstore/pull/420 until we find a better approach.
Putting an upper bound on `uv_build` causes publishing to our internal PyPI to fail (https://github.com/getsentry/pypi/pull/2117) because it only has `uv` versions `0.9.26` and `0.9.28` available in that index.


szokeasaurusrex (approved):
stamp

PR #463, opened 2026-05-07
feat(server): Add multipart upload endpoints

lcian:
⚠️ Stacked on https://github.com/getsentry/objectstore/pull/458

Final part of the Multipart uploads implementation server-side, this implements the endpoints on top of the Service.

Close FS-341
Close FS-359

### DIFF
--- a/objectstore-server/src/auth/service.rs
+++ b/objectstore-server/src/auth/service.rs
@@ -178,7 +178,7 @@ impl AuthAwareService {
         id: ObjectId,
         upload_id: UploadId,
     ) -> ApiResult<AbortMultipartResponse> {
-        self.assert_authorized(Permission::ObjectDelete, id.context())?;
+        self.assert_authorized(Permission::ObjectWrite, id.context())?;
         Ok(self.service.abort_multipart(id, upload_id).await?)
     }
 
--- a/objectstore-server/src/auth/service.rs
+++ b/objectstore-server/src/auth/service.rs
@@ -165,7 +165,7 @@ impl AuthAwareService {
         max_parts: Option<u32>,
         part_number_marker: Option<PartNumber>,
     ) -> ApiResult<ListPartsResponse> {
-        self.assert_authorized(Permission::ObjectRead, id.context())?;
+        self.assert_authorized(Permission::ObjectWrite, id.context())?;
         Ok(self
             .service
             .list_parts(id, upload_id, max_parts, part_number_marker)
--- a/objectstore-server/src/auth/service.rs
+++ b/objectstore-server/src/auth/service.rs
@@ -1,4 +1,8 @@
 use objectstore_service::id::{ObjectContext, ObjectId};
+use objectstore_service::multipart::{
+    AbortMultipartResponse, CompleteMultipartResponse, CompletedPart, InitiateMultipartResponse,
+    ListPartsResponse, PartNumber, UploadId, UploadPartResponse,
+};
 use objectstore_service::service::{DeleteResponse, GetResponse, InsertResponse, MetadataResponse};
 use objectstore_service::{ClientStream, StorageService};
 use objectstore_types::auth::Permission;
@@ -115,5 +119,80 @@ impl AuthAwareService {
     pub async fn delete_object(&self, id: ObjectId) -> ApiResult<DeleteResponse> {
         self.assert_authorized(Permission::ObjectDelete, id.context())?;
         Ok(self.service.delete_object(id).await?)
+    }
+
+    // --- Multipart upload operations ---
+
+    /// Auth-aware wrapper around [`StorageService::initiate_multipart`].
+    pub async fn initiate_multipart(
+        &self,
+        id: ObjectId,
+        metadata: Metadata,
+    ) -> ApiResult<InitiateMultipartResponse> {
+        self.assert_authorized(Permission::ObjectWrite, id.context())?;
+        Ok(self.service.initiate_multipart(id, metadata).await?)
+    }
+
+    /// Auth-aware wrapper around [`StorageService::upload_part`].
+    pub async fn upload_part(
+        &self,
+        id: ObjectId,
+        upload_id: UploadId,
+        part_number: PartNumber,
+        content_length: u64,
+        content_md5: Option<String>,
+        body: ClientStream,
+    ) -> ApiResult<UploadPartResponse> {
+        self.assert_authorized(Permission::ObjectWrite, id.context())?;
+        Ok(self
+            .service
+            .upload_part(
+                id,
+                upload_id,
+                part_number,
+                content_length,
+                content_md5,
+                body,
+            )
+            .await?)
+    }
+
+    /// Auth-aware wrapper around [`StorageService::list_parts`].
+    pub async fn list_parts(
+        &self,
+        id: ObjectId,
+        upload_id: UploadId,
+        max_parts: Option<u32>,
+        part_number_marker: Option<PartNumber>,
+    ) -> ApiResult<ListPartsResponse> {
+        self.assert_authorized(Permission::ObjectRead, id.context())?;
+        Ok(self
+            .service
+            .list_parts(id, upload_id, max_parts, part_number_marker)
+            .await?)
+    }
+
+    /// Auth-aware wrapper around [`StorageService::abort_multipart`].
+    pub async fn abort_multipart(
+        &self,
+        id: ObjectId,
+        upload_id: UploadId,
+    ) -> ApiResult<AbortMultipartResponse> {
+        self.assert_authorized(Permission::ObjectDelete, id.context())?;
+        Ok(self.service.abort_multipart(id, upload_id).await?)
+    }
+
+    /// Auth-aware wrapper around [`StorageService::complete_multipart`].
+    pub async fn complete_multipart(
+        &self,
+        id: ObjectId,
+        upload_id: UploadId,
+        parts: Vec<CompletedPart>,
+    ) -> ApiResult<CompleteMultipartResponse> {
+        self.assert_authorized(Permission::ObjectWrite, id.context())?;
+        Ok(self
+            .service
+            .complete_multipart(id, upload_id, parts)
+            .await?)
     }
 }
--- a/objectstore-server/src/endpoints/common.rs
+++ b/objectstore-server/src/endpoints/common.rs
@@ -90,6 +90,7 @@ impl ApiError {
             ApiError::Service(ServiceError::Client(_)) => StatusCode::BAD_REQUEST,
             ApiError::Service(ServiceError::Metadata(_)) => StatusCode::BAD_REQUEST,
             ApiError::Service(ServiceError::AtCapacity) => StatusCode::TOO_MANY_REQUESTS,
+            ApiError::Service(ServiceError::NotImplemented) => StatusCode::NOT_IMPLEMENTED,
             ApiError::Service(_) => {
                 objectstore_log::error!(!!self, "error handling request");
                 StatusCode::INTERNAL_SERVER_ERROR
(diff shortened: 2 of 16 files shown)


lcian:
bugbot run

lcian:
@sentry review

lcian:
bugbot run

lcian:
@sentry review

lcian:
bugbot run

lcian:
bugbot run

lcian:
@sentry review

objectstore-service/src/backend/common.rs
```diff
@@ -62,6 +62,15 @@ pub trait Backend: fmt::Debug + Send + Sync + 'static {
     /// (such as [`TieredStorage`](super::tiered::TieredStorage)) should override this
     /// to wait for those tasks to complete.
     async fn join(&self) {}
+
+    /// Returns a reference to the [`MultipartUploadBackend`] implementation,
+    /// if this backend supports multipart uploads.
+    ///
+    /// The default returns `None`. Backends that implement
+    /// [`MultipartUploadBackend`] should override this to return `Some(self)`.
+    fn as_multipart_upload_backend(&self) -> Option<&dyn MultipartUploadBackend> {
```
lcian: Not sure if this is the best approach.
`Backend` should not know about `MultipartUploadBackend`, and this means that the server errors with HTTP status NOT IMPLEMENTED when the backend doesn't support this capability.

As an alternative, we could make `StorageService` store a `Arc<dyn MultipartUploadBackend>`.
That however would prevent someone from using Objectstore with a Bigtable only configuration, which I'm not sure if we consider a legitimate usecase or not.

Regardless, we can easily make that change at any point.

objectstore-server/src/endpoints/multipart.rs
```diff
@@ -0,0 +1,339 @@
...
+}
+
+async fn initiate_inner(
+    service: AuthAwareService,
+    State(state): State<ServiceState>,
+    id: ObjectId,
+    headers: HeaderMap,
+) -> ApiResult<Response> {
+    let mut metadata = Metadata::from_headers(&headers, "").map_err(ServiceError::from)?;
+    // TODO: Do this in `complete` instead, when we have a Service API to mutate metadata.
```
lcian, line 153: This is easy to do but we would need a new method on `Backend`, plus all the implementations and plumbing through `Service`, filed a follow-up: https://linear.app/getsentry/issue/FS-356/set-correct-metadatatime-created-for-multipart-uploads.

Or, we can just accept and document that `time_created` refers to the time the upload was initiated, but that sounds wrong, it should probably be the time the logical object is created, which is the time the multipart upload is completed.

objectstore-server/src/auth/service.rs
```diff
@@ -116,4 +120,79 @@ impl AuthAwareService {
...
+
+    /// Auth-aware wrapper around [`StorageService::list_parts`].
+    pub async fn list_parts(
+        &self,
+        id: ObjectId,
+        upload_id: UploadId,
+        max_parts: Option<u32>,
+        part_number_marker: Option<PartNumber>,
+    ) -> ApiResult<ListPartsResponse> {
+        self.assert_authorized(Permission::ObjectRead, id.context())?;
```
lcian: Maybe this should also require `Permission::ObjectWrite` instead, to match the rest of the methods.
But also, if a user has `ObjectWrite` they would usually also have `ObjectRead` in any realistic scenario.

objectstore-server/src/endpoints/multipart.rs
```diff
@@ -0,0 +1,339 @@
...
+    Query(params): Query<UploadIdQuery>,
+    Json(body): Json<CompleteRequest>,
+) -> ApiResult<Response> {
+    service.check_permission(Permission::ObjectWrite, id.context())?;
+    for part in &body.parts {
+        validate_part_number(part.part_number)?;
+    }
+
+    let key = id.key().to_string();
+
```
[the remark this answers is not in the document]
lcian: Intended.

objectstore-server/src/auth/service.rs
matt-codecov: not sure what to do here. [GCS just gives listing its own permission](https://docs.cloud.google.com/storage/docs/access-control/iam-permissions#multipart-uploads)

> if a user has ObjectWrite they would usually also have ObjectRead in any realistic scenario

that might be how we assign permissions today, but i think probably most tokens only need to be read _or_ write.

if the point is to allow an interrupted write to continue, maybe write permissions should allow it? so a write-only client doesn't need to request additional permissions. idk

objectstore-server/src/endpoints/common.rs
```diff
@@ -90,10 +94,16 @@ impl ApiError {
...
             ApiError::Service(ServiceError::Metadata(_)) => StatusCode::BAD_REQUEST,
             ApiError::Service(ServiceError::AtCapacity) => StatusCode::TOO_MANY_REQUESTS,
+            ApiError::Service(ServiceError::NotImplemented) => StatusCode::NOT_IMPLEMENTED,
             ApiError::Service(_) => {
                 objectstore_log::error!(!!self, "error handling request");
                 StatusCode::INTERNAL_SERVER_ERROR
             }
+
+            ApiError::Internal(_) => {
+                objectstore_log::error!(!!self, "internal error");
```
matt-codecov, line 104: are we dropping important errors from these logs?

objectstore-service/src/error.rs
```diff
@@ -142,6 +146,7 @@ impl Error {
             Self::Panic(_) => Level::ERROR,
             Self::Dropped => Level::ERROR,
             Self::UnexpectedTombstone => Level::ERROR,
+            Self::NotImplemented => Level::WARN,
```
matt-codecov: this is probably an error? if this ever happens in a deployment it means something was misconfigured

objectstore-service/src/service.rs
```diff
@@ -233,6 +237,126 @@ impl StorageService {
...
+        let inner = Arc::clone(&self.inner);
+        self.spawn("initiate_multipart", async move {
+            inner
+                .as_multipart_upload_backend()
+                .unwrap()
+                .initiate_multipart(&id, &metadata)
+                .await
+        })
+        .await
+    }
```
matt-codecov: there's a way to streamline this a bit to get rid of the `ensure_...` helper and `.unwrap()` calls:

```rust
pub trait Backend {
  /// Make this an explicit `Arc<dyn T>` -> `Arc<dyn U>` cast.
  /// Return `Err(Error::NotImplemented)` by default.
  fn as_multipart_upload_backend(self: Arc<Self>) -> Result<Arc<dyn MultipartUploadBackend>> {
    Err(Error::NotImplemented)
  }
}

impl Backend for GcsBackend {
  /// Convert `Arc<dyn Backend>` into `Arc<dyn MultipartUploadBackend>`
  fn as_multipart_upload_backend(self: Arc<Self>) -> Result<Arc<dyn MultipartUploadBackend>> {
    Ok(self)
  }
}

impl StorageService {
  pub async fn initiate_multipart(...) -> Result<InitiateMultipartResponse> {
    // Clone `inner` so it can be `move`d into the async block, as before.
    // `.as_multipart_upload_backend()` transforms the clone into `Arc<dyn MultipartUploadBackend>`.
    // It still shares its underlying pointer/refcount with the original, though!
    // `?` to return `Err(Error::NotImplemented)` when not implemented.
    let inner_multipart = self.inner.clone().as_multipart_upload_backend()?;
    self.spawn("initiate_multipart", async move {
      inner_multipart.initiate_multipart(&id, &metadata).await
    }).await
  }
}
```

the `Arc` refcounting all still tracks the same object. this is just a super verbose cast for trait objects inside `Arc`s. you can confirm the refcount stuff in this playground link https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=ea6b827dd6fb9c3194bac54d0f62b2ea

objectstore-server/src/endpoints/multipart.rs
```diff
@@ -0,0 +1,339 @@
...
+struct CompleteErrorResponse {
+    error: CompleteErrorDetail,
+}
+
+fn validate_part_number(part_number: u32) -> ApiResult<()> {
+    if part_number == 0 {
+        return Err(ApiError::Client("part_number must be >= 1".into()));
+    }
+    Ok(())
+}
```
matt-codecov: would using a `type PartNumber = std::num::NonZero<u32>` alias instead of `u32` for these `part_number` fields take care of this? we'd have to move the error handling somewhere else but we wouldn't have to remember to call `validate()`

objectstore-server/src/endpoints/multipart.rs
```diff
@@ -0,0 +1,339 @@
...
+    service: AuthAwareService,
+    Xt(id): Xt<ObjectId>,
+    Query(params): Query<ListPartsQuery>,
+) -> ApiResult<Response> {
+    let response = service
+        .list_parts(
+            id,
+            params.upload_id,
+            params.max_parts,
+            params.part_number_marker,
```
matt-codecov, line 215: technically you have to validate that this is >=1 as well

objectstore-server/src/endpoints/multipart.rs
```diff
@@ -0,0 +1,339 @@
...
+    Ok(StatusCode::NO_CONTENT.into_response())
+}
+
+async fn complete(
+    service: AuthAwareService,
+    Xt(id): Xt<ObjectId>,
+    Query(params): Query<UploadIdQuery>,
+    Json(body): Json<CompleteRequest>,
+) -> ApiResult<Response> {
+    service.check_permission(Permission::ObjectWrite, id.context())?;
```
matt-codecov: is there something we need to check that isn't covered in `AuthAwareService`'s wrapper around `complete_multipart()`?

matt-codecov (approved):
accept w comments

objectstore-server/src/auth/service.rs
lcian: > that might be how we assign permissions today, but i think probably most tokens only need to be read or write.
> if the point is to allow an interrupted write to continue, maybe write permissions should allow it? so a write-only client doesn't need to request additional permissions. idk

This makes sense. 
At the end of the day, the only reason why you would `list_parts` is to eventually `complete` the upload, performing a write operation.
Therefore, I will change `list_parts` to require `ObjectWrite`.
Note that, even if you can list parts, there's no way to read the contents of an already uploaded part individually, so it's not like this would allow a user that has `ObjectWrite` to **read** part of an object or anything like that.

objectstore-server/src/endpoints/common.rs
lcian: Good catch. In this case we log `self` which is the complete `ApiError`. Sentry then reports it with its debug implementation and the complete `cause`/`source` chain, so we're not losing anything.

objectstore-service/src/service.rs
lcian: Very nice, thank you!!

objectstore-server/src/endpoints/multipart.rs
lcian: Good catch, removed.

objectstore-server/src/endpoints/multipart.rs
lcian: Very good suggestion, we're now using `NonZeroU32` for `part_number` and `part_number_marker` everywhere.

objectstore-server/src/endpoints/multipart.rs
lcian: See https://github.com/getsentry/objectstore/pull/463#discussion_r3311051775

lcian:
bugbot run

lcian:
@sentry review

objectstore-server/src/endpoints/multipart.rs
```diff
@@ -0,0 +1,324 @@
...
+    last_modified: SystemTime,
+    size: u64,
+}
+
+#[derive(Debug, Serialize)]
+struct ListPartsResponse {
+    parts: BTreeMap<PartNumber, PartInfo>,
+    is_truncated: bool,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    next_part_number_marker: Option<PartNumber>,
```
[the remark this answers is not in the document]
lcian: False.

PR #468, opened 2026-05-08
feat(rust-client): Add Multipart Upload API

lcian:
⚠️ Stacked on https://github.com/getsentry/objectstore/pull/463

Implements a multipart upload API for the Rust client.
Instead of going for a low-level API where calls match exactly HTTP calls as I've planned initially, I've decided to abstract some things, for example, initiating a multipart upload returns a `MultipartUpload` that represents it, much like the existing `Session` we have.
Please look at the README to see the intended usage.

Also makes a couple changes that aren't strictly related to the multipart upload impl:
- Moves multipart upload-related types from `objectstore-server` to `objectstore-types`, to make them usable from the client too, given that they both deal with the wire format.
- Makes the server return a vec of completed parts, each with their own `part_number` when calling `ListParts`, as opposed to a `BTreeMap`.
- Makes `Usecase::compression` take an `Option<Compression>`. This is a breaking change that makes it possible to disable default-compression on a usecase, which wasn't possible before.

Close FS-357


lcian:
bugbot run

lcian:
@sentry review

lcian:
bugbot run

lcian:
@sentry review

clients/rust/src/multipart.rs
```diff
@@ -0,0 +1,378 @@
...
+    metadata: Metadata,
+    key: Option<ObjectKey>,
+}
+
+impl InitiateMultipartBuilder {
+    /// Sets an explicit object key.
+    ///
+    /// If a key is specified, the object will be stored under that key. Otherwise, the Objectstore
+    /// server will automatically assign a random key, which is then returned from this request.
+    pub fn key(mut self, key: impl Into<ObjectKey>) -> Self {
```
lcian, line 91: Most of these methods are exactly the same as `PutBuilder`, except for `compression` which needs to have a separate docstring due to different behavior.
Should we unify the definition of the shared methods via a macro?

clients/rust/src/multipart.rs
```diff
@@ -0,0 +1,378 @@
...
+
+        loop {
+            let page = self.list_parts_page(None, marker).await?;
+            all_parts.extend(page.parts);
+
+            if !page.is_truncated {
+                return Ok(all_parts);
+            }
+            marker = page.next_part_number_marker;
+        }
```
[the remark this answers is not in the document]
lcian: This will never happen unless there's a bug in the server.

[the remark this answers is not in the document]
lcian: Also added a safeguard.

clients/rust/src/get.rs
```diff
@@ -128,7 +128,9 @@ pub(crate) fn maybe_decompress(
     match (metadata.compression, decompress && !encoding_accepted) {
         (Some(Compression::Zstd), true) => {
             metadata.compression = None;
-            ReaderStream::new(ZstdDecoder::new(StreamReader::new(stream))).boxed()
+            let mut decoder = ZstdDecoder::new(StreamReader::new(stream));
+            decoder.multiple_members(true);
```
matt-codecov, line 135: https://docs.rs/async-compression/latest/async_compression/tokio/bufread/struct.ZstdDecoder.html#method.multiple_members

not immediately apparent why you made this change?

clients/rust/src/multipart.rs
```diff
@@ -0,0 +1,384 @@
...
+    /// Creates a builder for initiating a multipart upload.
+    ///
+    /// The returned [`InitiateMultipartBuilder`] inherits the session's default compression
+    /// and expiration settings.
+    ///
+    /// IMPORTANT: unlike single-object uploads, the client does not automatically compress the
+    /// contents of [`MultipartUpload::put`]/[`MultipartUpload::put_stream`] based on the
+    /// configured `compression`.
+    /// The caller is responsible to compress the payload in accordance with the configured
+    /// `compression`.
```
matt-codecov, line 42: this also means they have to manually decompress, right?

clients/rust/src/multipart.rs
```diff
@@ -0,0 +1,384 @@
...
+            session: self.clone(),
+            metadata,
+            key: None,
+        }
+    }
+
+    /// Resumes an existing multipart upload from its key and upload ID.
+    ///
+    /// This reconstructs a [`MultipartUpload`] handle from previously obtained identifiers, and
+    /// doesn't make any network calls.
```
matt-codecov, line 63: it'll need to get the previously-uploaded parts to call `complete`, right?

clients/rust/src/multipart.rs
matt-codecov: we could maybe consolidate into an `UploadBuilder` type?
```rust
pub enum UploadBody {
    PutBody(...),
    MultipartUpload,
}
pub enum UploadResponse {
    PutResponse(...),
    MultipartResponse(...),
}
pub struct UploadBuilder {
    body: UploadBody,
}
impl UploadBuilder {
    pub fn send(...) -> Result<UploadResponse> {
        let mut request_builder = match self.body {
            UploadBody::PutBody(body) => self.request(...),
            UploadBody::MultipartUpload => self.multipart_request(...),
        };
        request_builder = request_builder.headers(...);
        request_builder.send().await.blah_blah_blah()?
    }
}
```

if we can't consolidate the type cleanly i would just leave it duplicated for now. we can macro it if we need to change the api often or duplicate it more

matt-codecov (commented):
ran out of steam before i could finish reviewing, sorry

clients/rust/src/multipart.rs
lcian: Nope, see the reply to https://github.com/getsentry/objectstore/pull/468#discussion_r3285214436.

clients/rust/src/get.rs
lcian: When the upload is done via a multipart upload and the object is compressed, the blob will consist of multiple zstd compressed blobs, concatenated to each other.
Using `multiple_members` here makes it possible for our `objectstore_client` to transparently decompress such format.
This means that the user can always use `session.get` and benefit from automatic decompression, regardless of how the object was uploaded.

lcian: I'll add a code comment explaining this.

lcian: I've checked and this is supported also for e.g. `gzip` if some day we want to support it.

clients/rust/src/multipart.rs
matt-codecov: just saw the README addition where the user is expected to do this. i guess that's okay - they might know that some part number is supposed to replace another part number, or something.

matt-codecov (approved):
accept w comments (including my comments last week)

clients/rust/src/multipart.rs
lcian: I don't really like making the return type an `UploadResponse` enum. Leaving it as is for now.

PR #477, opened 2026-05-13
feat: Add batch TOUCH operation with optimized implementation

lcian:
This adds a new TOUCH operation that, most importantly, can be batched.
Semantics: if an object exists, and has a TTI expiration policy, refresh its TTI.

The `Backend` trait gains a new method `touch_batch` that can provide an optimized implementation for executing multiple TOUCH operations. The default implementation just calls `get_metadata` in a loop and converts the result.
The GCS backend has an "optimized" implementation that executes the requests to GCS with bounded concurrency (20).

`HighVolumeBackend` also gains a new method `get_tiered_metadata_batch` with a similar default implementation.
The Bigtable backend has an optimized version where only two requests are used, a `ReadRows` to get the rows and a `MutateRows` to bump the TTI on all of them in bulk.

The name is inspired by the UNIX `touch` command, which bumps the date a file was last modified, although it's also different because it can also create files, which our operation doesn't. 
More discussion on the naming is welcome. 
The previous name was EXISTS, which IMO didn't really make it clear that this bumps TTI, so TOUCH might make that a bit clearer.

As a bonus, this also adds a HEAD API to the client, needed because with the way we have implemented batching, each batchable operation also needs its own non-batch implementation, and for TOUCH this is simply a HEAD request where we discard the metadata.


clients/rust/src/many.rs
```diff
@@ -13,10 +13,11 @@ use reqwest::header::{CONTENT_TYPE, HeaderMap, HeaderName, HeaderValue};
 use reqwest::multipart::Part;
```
lcian, line 13: This is all just plumbing.

lcian:
Unfortunately this can create internal inconsistency issues as we're using `MutateRows` and we could override a new tombstone update, leaving the new LT object orphaned.
Recreating a dumber version of this now.

PR #478, opened 2026-05-15
feat(server/rust-client): Add batchable HEAD operation

lcian:
This adds the HEAD operation to the Rust client API, and makes it batchable.

The intent is to use this in `sentry-cli` to do a batch check for existing Snapshots, bumping their TTI, and subsequently only upload the missing ones. This is possible given that we use the hash of the object as its key for snapshots.
Note that we will also need to change the expiration policy for Snapshots to TTI instead of TTL (currently used).

A singular HEAD request maps to a HEAD HTTP request.
A batch HEAD request goes through the existing `objectstore-service/src/streaming.rs` pipeline that executes operations with concurrency bounded to 10% of the existing service concurrency limit.

This is the "dumb" version of the much more optimized https://github.com/getsentry/objectstore/pull/477.
In that PR, I optimized the implementation down to the backend level, most importantly using `ReadRows` (plural) and `WriteRows` for Bigtable.
Unfortunately, that creates some possible inconsistency scenarios where, ultimately, LT blobs could remain orphaned. `WriteRows` is not conditional, so it could overwrite a new tombstone that comes in between the `ReadRows` to read all tombstones and the `WriteRows` to bump the TTI of existing ones.

Close FS-335


lcian:
bugbot run

lcian:
@sentry review

matt-codecov (approved):

PR #480, opened 2026-05-26
feat(changelog): Add and use Assembling phase with deferred cleanup

lcian:
⚠️ Stacked on https://github.com/getsentry/objectstore/pull/458

This addresses the issue being discussed in https://github.com/getsentry/objectstore/pull/458#discussion_r3200217052/ by introducing a new `ChangePhase::Assembling` variant.
The corresponding `Change` carries a `cleanup_after` timestamp, allowing the caller to defer cleanup.
In practice, we use this in `TieredStorage::complete_multipart` with an arbitrary 24 hours deadline (should we change this?), during which the user can freely retry completing the upload.

Note that this `ChangePhase` is special, as it behaves like `ChangePhase::complete` in-process:
- It does nothing on `Drop`, as we assume that the deadline hasn't passed yet.
- The corresponding `ChangeGuard` for it also doesn't spawn any task on `Drop`, due to the same assumption.

As a consequence, this depends on a persistent implementation of the `ChangeLog` to actually clean anything up, or at least on a task that periodically scans the in-memory `ChangeLog` for `Change`s to clean up, both of which we don't have yet.
In production, we will eventually always use a persistent `ChangeLog`, and other code paths exist that similarly require a persistent `ChangeLog` to actually ensure that no LT orphans are created, so I think it's fine for this change to also make that assumption.

Part of FS-339.
Will be squash-merged into https://github.com/getsentry/objectstore/pull/458 once approved.


lcian:
bugbot run

lcian:
@sentry review

objectstore-service/src/backend/tiered.rs
```diff
@@ -684,10 +699,11 @@ impl MultipartUploadBackend for TieredStorage {
...
         };
         let mut guard = self
-            .record_change(Change {
+            .record_assembling(Change {
                 id: id.clone(),
                 new: Some(physical.clone()),
                 old: current.clone(),
+                cleanup_after: Some(SystemTime::now() + MULTIPART_COMPLETE_CLEANUP_DELAY),
             })
             .await?;
```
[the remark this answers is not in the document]
lcian: This is intended, we need to record the `Change` at this exact point to ensure no orphans.

objectstore-service/src/backend/tiered.rs
```diff
@@ -684,10 +700,11 @@ impl MultipartUploadBackend for TieredStorage {
...
```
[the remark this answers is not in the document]
lcian: This is fine, we need to record the `Change` at this exact point to ensure no orphans are created.
Then, if `self.lt.complete_multipart` fails:
- the user will retry and `complete_multipart` will go through, causing the cleanup to be a no-op
- or, the user won't retry, and cleanup will also no-op as the blob doesn't exist

objectstore-service/src/backend/changelog.rs
```diff
@@ -373,12 +421,15 @@ impl ChangeState {
...
 impl Drop for ChangeState {
     fn drop(&mut self) {
-        if self.phase != ChangePhase::Completed {
-            objectstore_log::error!(
-                change = ?self.change,
-                phase = ?self.phase,
-                "Operation dropped without completing cleanup"
-            );
+        match self.phase {
+            ChangePhase::Assembling | ChangePhase::Completed => {}
```
matt-codecov: we probably do still want to log when a drop occurs in this phase

objectstore-service/src/backend/tiered.rs
matt-codecov: wasn't the goal to address a potential failure here? with your change, a failure here will still leave the change in the `Written` state which will result in the new LT blob being deleted and requiring the user to redo the whole thing.

should `Assembling` instead replace the `Written` state for multipart uploads?

lcian: Good catch, yeah my initial intention was for multipart uploads to hit the `Assembling` state instead of `Written`. Then we should go directly to a "final" state. Fixing.

lcian:
Okay it should be good now, here's the diff since the time you reviewed @matt-codecov https://github.com/getsentry/objectstore/pull/480/changes/6f626efafa4f2370cf0b227620eaaab71ef3c99f..d32e4c2dbda9e74858d9757bcc23fdde0c14d4e4

objectstore-service/src/backend/tiered.rs
```diff
@@ -672,47 +688,87 @@ impl MultipartUploadBackend for TieredStorage {
...
 
-        // 1. Read current HV revision to establish the write precondition
+        let physical = ObjectId {
+            context: id.context.clone(),
+            key: tiered.revision,
+        };
+
+        // 1. Read current HV revision to establish the write precondition.
         let current = match self.inner.high_volume.get_tiered_metadata(id).await? {
+            // Optimization: a previous attempt already finalized this revision and tombstone -- report success.
```
lcian, line 698: Not sure if we want to keep it but I guess it doesn't hurt?

matt-codecov (approved):

PR #486, opened 2026-06-03
fix(server): Apply Sentry middleware first

lcian:
I noticed that some of our errors in Sentry have wrong trace correlation, and e.g. logs for multiple requests show up in the logs tab for a single error (this is just a symptom of wrong trace correlation i.e. using the wrong Hub).

I haven't found the root cause, so this probably doesn't really fix anything, but it's the correct order in principle.

However, if we logged anything in the other middleware (which I think we do in `limit_web_concurrency`), without this fix those logs would not show up under the trace for the request we're making (because there wouldn't be one yet), so this fixes at least those potential cases.

### DIFF
--- a/objectstore-server/src/web/app.rs
+++ b/objectstore-server/src/web/app.rs
@@ -32,9 +32,9 @@ impl App {
         //  - Requests go from top to bottom
         //  - Responses go from bottom to top
         let middleware = ServiceBuilder::new()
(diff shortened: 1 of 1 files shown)


matt-codecov (approved):
maybe move these after the `emit_request_metrics` middleware? i think i am more concerned with sentry overhead being excluded from datadog metrics than i am with datadog overhead being excluded from sentry instrumentation. idk tho up to you

PR #487, opened 2026-06-03
feat!(types): Use MediaTypeBuf for Metadata.content_type

lcian:
I noticed that we were using `Cow<'static, str>` and then using a util `vaildate_content_type` when deserializing, which I think is kinda weird, so this changes the type to use `mediatype::MediaTypeBuf` directly.

Most paths still remain fallible due to the need to de/serialize this field.

This is a breaking change for the rust client, as users will now have to pass a `MediaTypeBuf`.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2139,6 +2139,9 @@ name = "mediatype"
 version = "0.21.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "120fa187be19d9962f0926633453784691731018a2bf936ddb4e29101b79c4a7"
+dependencies = [
+ "serde",
+]
 
 [[package]]
 name = "memchr"
@@ -2571,6 +2574,7 @@ dependencies = [
  "futures-util",
  "infer",
  "jsonwebtoken",
+ "mediatype",
  "multer",
  "objectstore-test",
  "objectstore-types",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -40,6 +40,7 @@ http = "1.3.1"
 humantime = "2.2.0"
 humantime-serde = "1.1.1"
 jsonwebtoken = { version = "10.3.0", features = ["rust_crypto"] }
+mediatype = { version = "0.21.0", features = ["serde"] }
 rand = "0.9.3"
 regex = "1.12.3"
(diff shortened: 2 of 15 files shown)


matt-codecov (changes requested):
i'm guessing it was done this way to keep the third-party `mediatype` crate out of our public API. that way, if the crate ever goes unmaintained or we adopt a different dep that includes the same functionality, we can switch without it being a breaking change / major version bump. i agree with that decision so i'm requesting changes to this PR

that said, i do think this is a bit weird. why is `metadata.content_type` of type `Cow<'static, str>`?
- could we just do `&'static str` or do we actually modify content types in-place somewhere in a way that triggers the clone? if so, should we just bite the bullet and make this a `String` from the jump?
- does the lifetime really need to be `'static`? what if a caller knows the content type without needing to scan the contents but the content type was read from a DB or a response from some other service and has a non-static lifetime?

lcian:
> - could we just do `&'static str` or do we actually modify content types in-place somewhere in a way that triggers the clone? if so, should we just bite the bullet and make this a `String` from the jump?
> - does the lifetime really need to be `'static`? what if a caller knows the content type without needing to scan the contents but the content type was read from a DB or a response from some other service and has a non-static lifetime?

I think that's exactly why we use `Cow<'static, str>`. 
It's not really about the clone on write, as we never need to write that field, but more about the fact that `Cow<'static, str>` acts essentially as an enum that can hold either a `&'static str` or an owned `String`, supporting both cases you mentioned and being slightly more optimized when the string is a constant.

lcian:
> i'm guessing it was done this way to keep the third-party mediatype crate out of our public API.

I think that's a good point, this PR was more of a nit to have stronger typing, but I realize that changing this would probably have more drawbacks than advantages.

We could keep it `Cow<'static, str>` in the client API but change it to `MediaType`/`MediaTypeBuf` only in `objectstore-types`, but that now means you will have to duplicate `content_type` in a bunch of places as you'll want to convert it when sending the request and bubble up the error, or you need to change the methods that set `content_type` on a request to return a `Result` which we've been trying to avoid in favor of only erroring when sending the request.

The only advantage this could provide is that if you supply an invalid `content_type` you would avoid making the HTTP roundtrip as you would fail client-side, but that's pretty unlikely anyway.

PR #491, opened 2026-06-05
feat(metrics): Add and use timer guard

lcian:
⚠️ Stacked on https://github.com/getsentry/objectstore/pull/485

Addresses the issues flagged in https://github.com/getsentry/objectstore/pull/485 by introducing `objectstore_metrics::timer!` that returns a `TimerGuard` which records elapsed time on drop or explicit consumption with `record()`. See code for API and more info.

Also updates all the existing usages of timing metrics to use of this new API.

Close FS-375


objectstore-metrics/src/lib.rs
```diff
@@ -82,6 +82,91 @@ impl AsF64 for std::time::Duration {
...
+
+    /// Adds a tag to the metric.
+    pub fn tag(mut self, key: &'static str, value: impl Into<metrics::SharedString>) -> Self {
+        self.labels.push(metrics::Label::new(key, value));
+        self
+    }
+
+    /// Changes the behavior of this guard to always record the metric
+    /// with `success:true`, even on drop.
+    pub fn success(mut self) -> Self {
```
lcian, line 130: We don't need this atm but I can see some cases where this would realistically be useful, e.g. on an idempotent function that shortcircuits as an optimization.

lcian:
bugbot run

lcian:
@sentry review

matt-codecov (approved):

PR #492, opened 2026-06-05
feat(cogs): objectstore.cogs.usage datadog metric

matt-codecov:
imitating how relay does compute COGS, except we are using request _counts_ rather than request _timings_ because we're network-bound and CPU-light

every request incrs the counter by 1, except batch requests incr it for each operation in the batch

need to update usage-accountant config to consume the counter but then compute COGS is finished

Ref FS-195

### DIFF
--- a/objectstore-service/src/backend/counting.rs
+++ b/objectstore-service/src/backend/counting.rs
@@ -1,0 +1,347 @@
+//! Defines [`CountingBackend`], a decorator for the [`Backend`] trait that emits Cost of Goods Sold
+//! (COGS) compute usage metrics to the `objectstore.cogs.usage` counter.
+//!
+//! [`CountingBackend`] is meant to wrap the outer-most [`Backend`] implementation owned by
+//! [`StorageService`] so that every tracked backend operation, whether a single-object operation
+//! called by [`StorageService`] or a batched operation streamed by [`StreamExecutor`], is counted
+//! once. Notably, any operation that fails before it gets to [`StorageService`] (e.g. an auth or
+//! rate limit failure at a higher layer) is not counted.
+//!
+//! For COGS purposes we use operation count as a proxy for compute cost under the assumption that
+//! each request we serve has a basically flat CPU cost. Large payloads take longer, but they can be
+//! streamed in the background while other requests are served so they don't really cost more.
+//!
+//! [`StorageService`]: crate::service::StorageService
+//! [`StreamExecutor`]: crate::streaming::StreamExecutor
+
+use std::sync::Arc;
+
+use objectstore_types::metadata::Metadata;
+use objectstore_types::range::ByteRange;
+
+use crate::backend::common::{
+    Backend, DeleteResponse, GetResponse, MetadataResponse, MultipartUploadBackend, PutResponse,
+};
+use crate::error::Result;
+use crate::id::ObjectId;
+use crate::multipart::{
+    AbortMultipartResponse, CompleteMultipartResponse, CompletedPart, InitiateMultipartResponse,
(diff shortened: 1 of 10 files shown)


objectstore-server/src/web/middleware.rs
```diff
@@ -162,14 +192,93 @@ mod tests {
...
+                .await
+                .unwrap();
+            assert_eq!(resp.status(), StatusCode::OK);
+        });
+
+        assert!(
+            !captured
+                .iter()
+                .any(|m| m.starts_with("objectstore.cogs.usage")),
+            "unexpected cogs usage metric for internal route, captured: {captured:?}"
```
lcian:
```suggestion
            "unexpected cogs usage metric for batch route, captured: {captured:?}"
```

objectstore-server/src/cogs.rs
```diff
@@ -0,0 +1,33 @@
+//! Cost of Goods Sold (COGS) instrumentation utilities for Objectstore.
+
+/// Converts an Objectstore `usecase` to an appropriate `app_feature` value for use in our COGS
+/// pipelines.
+fn usecase_to_appfeature(usecase: &str) -> &'static str {
+    // TODO: Flesh out this mapping
+    match usecase {
+        "attachments" => "attachments",
```
lcian: This query can give you the categories: https://redash.getsentry.net/queries/11424/source.
For snapshots I guess we should use `preprod`. 
I can also see there's `errors_attachments` and `standalone_attachments`, but from the POV of Objectstore we cannot distinguish between the two, so it's right to use `attachments`.

objectstore-server/src/endpoints/batch.rs
```diff
@@ -109,8 +112,17 @@ async fn batch(
...
-    // Step 5: stamp inserts with time_created and record bandwidth
-    let stamped = policy_checked.map({
+    // Step 5: Increment COGS counter for each operation
+    let cogs_counted = validate(policy_checked, {
+        let usecase = context.usecase.clone();
+        move |_op| {
+            cogs::count(&usecase);
+            Ok(())
+        }
+    });
```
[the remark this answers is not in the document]
lcian: Not sure if we want to take into account rejected ops (probably not?), but surely we would want to be consistent between batch/non-batch either way.

lcian (approved):
Please address comment about single vs batch and rejected vs accepted requests.

objectstore-service/src/backend/counting.rs
```diff
@@ -0,0 +1,353 @@
...
+/// - `inner: Arc<dyn Backend>`
+/// - `inner_multipart: Option<Arc<dyn MultipartUploadBackend>>` if `inner` supports it
+///
+/// [`CountingBackend`]'s implementation clashes with how the [`MultipartUploadBackend`] trait is
+/// connected to the [`Backend`] trait: [`CountingBackend::as_multipart_upload_backend`] can't apply
+/// the cast to `self.inner` which remains an `Arc<dyn Backend>`. There are two workarounds:
+/// - (Rejected) Every multipart method must call `self.inner.clone().as_multipart_upload_backend()?`
+///   which clones every time
+/// - (Chosen) `CountingBackend` must call `self.inner.clone().as_multipart_upload_backend().ok()`
+///   and hang onto it
```
lcian: This is a different way to work around the problem with a minor disadvantage, but IMO preferable to this approach: https://github.com/getsentry/objectstore/pull/494

lcian (approved):
It seems sensible we do it by wrapping the Service, so that we count every operation that goes through there.
I would prefer if we work around the multipart thing using https://github.com/getsentry/objectstore/pull/494 instead of holding two `Backend`s.

PR #494, opened 2026-06-09
ref: Change as_multipart_upload_backend to borrow instead of Arc

lcian:


objectstore-service/src/service.rs
```diff
@@ -252,9 +252,12 @@ impl StorageService {
         id: ObjectId,
         metadata: Metadata,
     ) -> Result<InitiateMultipartResponse> {
-        let inner = self.inner.clone().as_multipart_upload_backend()?;
+        let inner = Arc::clone(&self.inner);
         self.spawn("initiate_multipart", async move {
-            inner.initiate_multipart(&id, &metadata).await
+            inner
+                .as_multipart_upload_backend()?
```
lcian, line 259: This is the only disadvantage, i.e. we will spawn a task unnecessarily upon every request if the backend doesn't support multipart uploads.

matt-codecov: actually, this lets us improve on the previous solution. we used to have to clone before we knew whether multipart was supported. now we don't. you can attempt the borrow without using the result like you did in `CountingBackend`

i'll incorporate this into my PR

PR #501, opened 2026-06-17
feat(server): Add jemalloc stats as Datadog metrics

jan-auer:
Objectstore shows RSS growth over time in production. Before investing in heap profiling, we need to distinguish real allocation growth from allocator fragmentation. This adds four jemalloc internal stats as Datadog gauges emitted on the existing `metrics_interval` (default 10s): `jemalloc.allocated`, `jemalloc.active`, `jemalloc.resident`, and `jemalloc.mapped`. Comparing `allocated` (what the app requested) against `resident` (what the OS reports) will make that distinction visible in production.

Also bumps `tikv-jemallocator` from 0.6.1 to 0.7.0 (jemalloc 5.3.1, no breaking changes) and adds `tikv-jemalloc-ctl 0.7.0` with the `stats` feature. Both must be the same version to link against the same `tikv-jemalloc-sys`.

### DIFF
--- a/objectstore-server/src/state.rs
+++ b/objectstore-server/src/state.rs
@@ -57,6 +57,7 @@ impl Services {
     /// use in the web server.
     pub async fn spawn(config: Config) -> Result<ServiceState> {
         tokio::spawn(track_runtime_metrics(config.runtime.metrics_interval));
+        tokio::spawn(track_allocator_metrics(config.runtime.metrics_interval));
 
         let backend = backend::from_config(config.storage.clone()).await?;
         let service =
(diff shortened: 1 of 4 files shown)


objectstore-server/src/state.rs
```diff
@@ -108,6 +109,43 @@ impl Services {
...
+    let allocated = tikv_jemalloc_ctl::stats::allocated::mib().expect("jemalloc allocated MIB");
+    let active = tikv_jemalloc_ctl::stats::active::mib().expect("jemalloc active MIB");
+    let resident = tikv_jemalloc_ctl::stats::resident::mib().expect("jemalloc resident MIB");
+    let mapped = tikv_jemalloc_ctl::stats::mapped::mib().expect("jemalloc mapped MIB");
+
+    let mut ticker = tokio::time::interval(interval);
+    loop {
+        ticker.tick().await;
+
+        let Ok(_) = epoch.advance() else {
```
lcian, line 126: What does this do?

lcian (approved):

objectstore-server/src/state.rs
jan-auer: This refreshes stats, which may otherwise be cached. As per docs of the crate, this should be called when reading stats in a loop like we do.

objectstore-server/src/state.rs
```diff
@@ -108,6 +109,43 @@ impl Services {
     }
 }
 
+/// Periodically captures and reports jemalloc stats.
+async fn track_allocator_metrics(interval: Duration) {
+    // INVARIANT: MIB resolution only fails if jemalloc is not the active allocator,
+    // which would be a misconfigured build. Panic early to surface the problem.
+    let epoch = tikv_jemalloc_ctl::epoch::mib().expect("jemalloc epoch MIB");
```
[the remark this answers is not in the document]
jan-auer: For allocator metrics, this is fine. We will still see the error.

PR #506, opened 2026-06-18
fix(service): Honor byte range requests in Bigtable backend

jan-auer:
The Bigtable backend ignored the `ByteRange` parameter in `get_tiered_object` (the parameter was named `_range`), returning the full object with no `ContentRange` regardless of what the client requested. This meant clients always received 200 with the complete payload instead of 206 Partial Content.

Adds an `apply_range` helper that resolves the byte range against the payload and produces a zero-copy `Bytes::slice`, then wires it into the `get_tiered_object` return path.


lcian (approved):
LGTM

PR #507, opened 2026-06-18
build: Consolidate workspace dependencies and pin explicit versions

jan-auer:
- Replace `tokio` `features = ["full"]` with the 7 features objectstore-server actually uses (`fs`, `macros`, `net`, `rt-multi-thread`, `signal`, `sync`, `time`)
- Consolidate 28+ direct member-crate dependencies into `[workspace.dependencies]` as a single alphabetically-sorted block
- Pin all abbreviated version specifiers (`x.y` → `x.y.z`) and align with Cargo.lock resolved versions
- Remove unused `sentry-options` workspace dependency

### DIFF
--- a/clients/rust/Cargo.toml
+++ b/clients/rust/Cargo.toml
@@ -11,7 +11,7 @@ rust-version = "1.89"
 publish = true
 
 [dependencies]
-async-compression = { workspace = true }
+async-compression = { workspace = true, features = ["tokio", "zstd"] }
 base64 = { workspace = true, optional = true }
 bytes = { workspace = true }
 futures-util = { workspace = true }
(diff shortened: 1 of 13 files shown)


clients/rust/Cargo.toml
```diff
@@ -11,24 +11,25 @@ rust-version = "1.89"
...
+infer = { workspace = true }
 jsonwebtoken = { workspace = true }
-multer = "3.1.0"
+multer = { workspace = true }
 objectstore-types = { workspace = true }
+percent-encoding = { workspace = true }
+# Pinned below workspace version for MSRV compatibility
 reqwest = { version = "0.13.1", default-features = false, features = ["charset", "http2", "system-proxy", "json", "stream", "multipart"] }
-sentry-core = { version = ">=0.41", features = ["client"] }
+sentry-core = { workspace = true, features = ["client"] }
```
lcian: This was intentionally specifying an open range so that the user could bring their own version and it would unify.
This way we're forcing users to either upgrade to `0.48.x`, or if they don't notice, they will have 2 different versions, which means that the application and our `objectstore-client` will have 2 separate SDK states, meaning that this is not gonna work as intended.
Granted, we're only using this to propagate tracing headers.

Also this can probably be `default-features = false` now that I see it.

lcian (approved):

PR #513, opened 2026-06-23
fix: Apply pedantic Clippy lints across workspace

jan-auer:
Applies a batch of pedantic Clippy lints across the workspace. Some occurrences were intentionally skipped where the suggested change would reduce readability or break forward compatibility (e.g. deriving `Eq` on types that may gain non-`Eq` fields).

Lints addressed:

- `uninlined_format_args` — inline format arguments
- `cloned_instead_of_copied` — use `.copied()` for `Copy` types
- `default_trait_access` — use `Type::default()` instead of `Default::default()`
- `single_char_pattern` — char literals instead of single-char string slices
- `explicit_into_iter_loop` — drop redundant `.into_iter()` in for loops
- `derive_partial_eq_without_eq` — add `Eq` when `PartialEq` is derived and all fields support it
- `redundant_closure_for_method_calls` — method references instead of closures
- `needless_qualified_path` — use already-imported names without full path prefix

### DIFF
--- a/clients/rust/src/client.rs
+++ b/clients/rust/src/client.rs
@@ -161,7 +161,7 @@ impl Usecase {
         Self {
             name: name.into(),
             compression: Some(Compression::Zstd),
-            expiration_policy: Default::default(),
+            expiration_policy: ExpirationPolicy::default(),
         }
     }
 
@@ -430,7 +430,7 @@ impl Session {
             .push("objects")
             .push(&self.scope.usecase.name)
             .push(&self.scope.scopes.as_api_path().to_string())
-            .extend(object_key.split("/"));
+            .extend(object_key.split('/'));
         drop(segments);
 
         url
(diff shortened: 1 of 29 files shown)


clients/rust/src/client.rs
```diff
@@ -161,7 +161,7 @@ impl Usecase {
         Self {
             name: name.into(),
             compression: Some(Compression::Zstd),
-            expiration_policy: Default::default(),
+            expiration_policy: ExpirationPolicy::default(),
```
lcian, line 164: I wonder why `Default::default` is not good here.

jan-auer: Using `Default::default()` is perfectly fine. This is mainly my personal preference because it is easier to recognize the type that is being constructed in contexts like GitHub where intellisense is not available. 

In this particular case, it's self-describing thanks to the field name. In other cases it can be helpful, for example `rules: Vec::new()` vs `rules: HashMap::new()`. 

The `default_trait_access` clippy lint is in the `pedantic` ruleset precisely for the reason that this is neither a correctness issue, nor is it objectively better or worse.

lcian (approved):
LGTM

PR #515, opened 2026-06-23
fix: Use jemalloc only on linux

jan-auer:
`tikv-jemallocator` was pulled in unconditionally on all platforms. On macOS, every startup printed:

```
<jemalloc>: option background_thread currently supports pthread only
```

jemalloc's background-thread purging requires pthread (Linux only); on macOS the option is silently ignored but the warning still fires. On Windows (MSVC), `jemalloc-sys` has no support at all and the build fails.

objectstore runs as a Linux service in production, so the fix is to gate both the Cargo dependencies and all code references behind `cfg(target_os = "linux")`, falling back to the system allocator on other platforms.


lcian (approved):
LGTM

PR #516, opened 2026-06-23
feat(client): Accept optional tokens in ClientBuilder::token

jan-auer:
The `Client` builder supports chaining methods. However, there was no way to conditionally supply a token generator while chaining builder methods, which is unergonomic.

`ClientBuilder::token` now accepts an `Option` directly via a new `IntoTokenProvider` trait. A `None` leaves the client unauthenticated. The new trait is also implemented for all types that were accepted previously, keeping backwards compatibility.

This brings the Rust client in line with the Python client, which already accepts an optional token:

```rust
// before
let mut builder = Client::builder(remote);
if let Some(token) = token_opt {
    builder = builder.token(token);
}
let client = builder.build()?;

// after
let client = Client::builder(remote)
    .token(token_opt)
    .build()?;
```

### DIFF
--- a/clients/rust/README.md
+++ b/clients/rust/README.md
@@ -286,6 +286,7 @@ via [`ClientBuilder::token`]. It accepts either:
   and scope being accessed.
 - A **`String` / `&str`** — a pre-signed JWT, used as-is for every request.
   Use this for external services that receive a token from another source.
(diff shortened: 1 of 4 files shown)


lcian (approved):
Looks good.

PR #517, opened 2026-06-23
feat: Add filename field to Metadata

lcian:
For Debug Files, the current Sentry endpoint returns a `Content-Disposition: attachment; filename="<filename>"` header, which prompts browsers to save the file under the given name.
Objectstore doesn't currently return this, so if we want downloads from e.g. the UI to go directly to Objectstore, the user will likely be prompted with the Objectstore key as the filename, which doesn't necessarily match the logical filename we show e.g. in the UI itself.

This introduces a first-class `Metadata` field to the server and clients, which is send by clients as `x-sn-filename` and rendered in GET/HEAD responses via such a `Content-Disposition` header.

This field can be an arbitrary `String`.
When deserializing it into a `Content-Disposition`, we perform escaping and normalization of path separators and filenames containing only dots.

Close FS-134


lcian:
bugbot run
@sentry review

objectstore-service/src/streaming.rs
```diff
@@ -84,7 +84,7 @@ pub struct Head {
 #[derive(Debug)]
 pub enum Operation {
     /// Insert a new object.
-    Insert(Insert),
+    Insert(Box<Insert>),
```
lcian, line 87: `insert.metadata` gained one field, so this was firing clippy's large enum variant lint.

lcian:
bugbot run

objectstore-types/src/metadata.rs
```diff
@@ -386,6 +404,24 @@ impl Metadata {
     }
 }
 
+/// Formats a `Content-Disposition: attachment` header value for the given filename.
+///
+/// Escapes `\` and `"` in the filename per RFC 6266 quoted-string rules.
+pub fn format_content_disposition(filename: &str) -> String {
```
jan-auer: Note that `axum-extra` would have utilities for this: https://docs.rs/axum-extra/latest/axum_extra/response/struct.Attachment.html

At least, we should compare with their implementation to ensure we're not missing something.

clients/python/tests/test_e2e.py
```diff
@@ -205,12 +206,13 @@ def test_full_cycle_with_origin(server_url: str) -> None:
 
     session = client.session(test_usecase, org=42, project=1337)
 
-    object_key = session.put(b"test data", origin="203.0.113.42")
+    object_key = session.put(b"test data", origin="203.0.113.42", filename="report.pdf")
```
jan-auer, line 209: could we add tests for header injection, i.e.:
- backslash and double quotes in the filename: need escaping in quoted headers (which we use)
- forward slash in the filename: would create a path and we likely don't want that
- `.` and `..` as filename: as above
- control characters (0x00–0x1F, 0x7F) in the filename: not permitted in header values by 7230

i currently can't think of other things we should reject. 

same for rust client. i think we should sanitize this in the client and validate server-side.

lcian: Updated the approach as we discussed offline, and added the requested tests.
Control characters are structurally impossible because the value itself needs to be turned into `HeaderValue` for transmission, both from the client to the server, as well as the other way around, so the respective party will just fail serializing.

objectstore-server/src/endpoints/objects.rs
```diff
@@ -125,10 +126,51 @@ async fn object_head(service: AuthAwareService, Xt(id): Xt<ObjectId>) -> ApiResu
     let headers = metadata.to_headers("").map_err(ServiceError::from)?;
 
     let mut response = (StatusCode::NO_CONTENT, headers).into_response();
+    insert_content_disposition(&mut response, &metadata);
```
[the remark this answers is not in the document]
lcian: Intended.

clients/python/src/objectstore_client/metadata.py
```diff
@@ -61,6 +62,14 @@ class Metadata:
     (e.g., the IP of a Sentry SDK or CLI).
     """
 
+    filename: str | None
```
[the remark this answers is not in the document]
lcian: That's fine.

clients/rust/src/put.rs
```diff
@@ -168,6 +168,16 @@ impl PutBuilder {
...
     }
 
+    /// Sets the filename of the object.
+    ///
+    /// When present, the server will include a `Content-Disposition: attachment; filename="<filename>"`
+    /// header in GET responses, prompting browsers and download tools to save the file under
+    /// this name.
+    pub fn filename(mut self, filename: impl Into<String>) -> Self {
+        self.metadata.filename = Some(filename.into());
+        self
```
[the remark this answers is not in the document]
lcian: Fixed.

jan-auer (approved):

PR #519, opened 2026-06-24
ref(server): Skip auth checks when not configured

jan-auer:
Auth checks previously ran on every request regardless of server-side configuration, producing noisy warnings in environments without auth (e.g. local development). Auth is now skipped entirely when it is not configured — meaning no keys are present and enforcement is off (`AuthZ::is_active()`).

When auth is active, `enforce` controls whether failures reject the request (debug log) or just warn. `AuthAwareService::new` is now infallible since the "enforcement enabled but no context" error state is no longer reachable.

Ref FS-344
Ref FS-313

### DIFF
--- a/objectstore-server/src/auth/error.rs
+++ b/objectstore-server/src/auth/error.rs
@@ -1,4 +1,3 @@
-use objectstore_types::auth::Permission;
 use thiserror::Error;
 
 /// Error type for different authorization failure scenarios.
@@ -39,18 +38,17 @@ impl AuthError {
         }
     }
 
-    /// Increment a counter and emit a debug log for this auth failure.
+    /// Increment a counter and emit a log for this auth failure.
     ///
-    /// If `enforce` is false, authentication failures will be logged as warnings to ensure they
-    /// are found and fixed to unblock enabling enforcement.
-    pub fn log(&self, permission: Option<Permission>, usecase: Option<&str>, enforce: bool) {
+    /// If `warn` is true, the log will be at WARN level; otherwise it will be at DEBUG level.`
+    pub fn log(&self, warn: bool) {
         let code = self.code();
         objectstore_metrics::count!("server.auth.failure", code = code);
-        let msg = self.to_string();
-        if !enforce {
-            objectstore_log::warn!(?permission, ?usecase, ?code, ?msg, "Auth failure");
+
+        if warn {
+            objectstore_log::warn!(code, reason=%self, "Auth failure");
         } else {
-            objectstore_log::debug!(?permission, ?usecase, ?code, ?msg, "Auth failure");
+            objectstore_log::debug!(code, reason=%self, "Auth failure");
         }
     }
 }
--- a/objectstore-server/src/auth/error.rs
+++ b/objectstore-server/src/auth/error.rs
@@ -21,7 +21,7 @@ pub enum AuthError {
     #[error("failed to verify token")]
     VerificationFailure,
 
-    /// Indicates that the requested operation is not authorized and auth enforcement is enabled.
+    /// Indicates that the requested operation is not permitted on the resource.
     #[error("operation not allowed")]
     NotPermitted,
 }
@@ -40,7 +40,7 @@ impl AuthError {
 
     /// Increment a counter and emit a log for this auth failure.
     ///
-    /// If `warn` is true, the log will be at WARN level; otherwise it will be at DEBUG level.`
+    /// If `warn` is true, the log will be at WARN level; otherwise it will be at DEBUG level.
     pub fn log(&self, warn: bool) {
         let code = self.code();
         objectstore_metrics::count!("server.auth.failure", code = code);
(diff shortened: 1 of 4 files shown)


objectstore-server/src/auth/error.rs
```diff
@@ -39,18 +38,17 @@ impl AuthError {
...
-    pub fn log(&self, permission: Option<Permission>, usecase: Option<&str>, enforce: bool) {
+    /// If `warn` is true, the log will be at WARN level; otherwise it will be at DEBUG level.`
+    pub fn log(&self, warn: bool) {
         let code = self.code();
         objectstore_metrics::count!("server.auth.failure", code = code);
-        let msg = self.to_string();
-        if !enforce {
-            objectstore_log::warn!(?permission, ?usecase, ?code, ?msg, "Auth failure");
+
+        if warn {
```
jan-auer, line 48: This moves the responsibility to decide on the log level to the caller, which actually carries the necessary info. It's a more appropriate place than on the error.

objectstore-server/src/auth/error.rs
```diff
@@ -39,18 +38,17 @@ impl AuthError {
...
+            objectstore_log::warn!(code, reason=%self, "Auth failure");
```
jan-auer, line 49: Permission and usecase were only passed at a single place where they are available. Thus, we're now configuring them on the scope, removing two special-case parameters from this function's signature. The values are still available in the Sentry error.

Note that they won't be printed to stderr anymore, which we have not needed.

objectstore-server/src/auth/service.rs
```diff
@@ -49,44 +49,32 @@ impl AuthAwareService {
...
-            Some(auth) => auth.assert_authorized(perm, context),
-            None => Ok(()),
-        }
-        .inspect_err(|err| err.log(Some(perm), Some(context.usecase.as_str()), self.enforce));
-
-        match self.enforce {
-            true => Ok(auth_result?),
-            false => Ok(()),
+    pub fn new(service: StorageService, context: Option<AuthContext>, enforce: bool) -> Self {
+        Self {
```
jan-auer, line 53: `enforce && context.is_none()` is no longer a valid state, but also the contract has changed:
 - the extractor still has the responsibility for validating this
 - `context: None` now is the primary argument and `enforce` only controls what happens _when_ a context is provided. 

The new doc comment above elaborates on this. Note that this is just the contract for `AuthAwareService`, not the public config!

objectstore-server/src/auth/service.rs
```diff
@@ -49,44 +49,32 @@ impl AuthAwareService {
...
 
     /// Checks whether the request is authorized for the given permission on the given context.
     ///
-    /// Returns `Ok(())` if authorized, or otherwise an error indicating the reason.
-    /// Equivalent to the internal `assert_authorized` check but exposed for callers
-    /// that validate operations individually before delegating to a lower-level service.
+    /// Returns `Ok(())` if authorized, or an error indicating the reason.
     pub fn check_permission(&self, perm: Permission, context: &ObjectContext) -> ApiResult<()> {
-        self.assert_authorized(perm, context)
+        if let Some(auth) = &self.context
```
jan-auer, line 64: This was inlined. Apparently, the `assert_authorized` method was intended to become public, but instead a plain wrapper created.

objectstore-server/src/extractors/service.rs
```diff
@@ -20,31 +20,30 @@ impl FromRequestParts<ServiceState> for AuthAwareService {
         parts: &mut Parts,
         state: &ServiceState,
     ) -> Result<Self, Self::Rejection> {
-        let encoded_token = parts
+        let enforce = state.config.auth.enforce;
+        if !state.config.auth.is_active() {
+            return Ok(AuthAwareService::new(state.service.clone(), None, enforce));
+        }
```
jan-auer, line 26: This is the main difference: If auth is inactive, we skip everything. 

Everything below is equivalent to before and only refactored for readability / brevity.

objectstore-server/src/config.rs
```diff
@@ -357,6 +357,17 @@ fn default_enforce() -> bool {
...
 }
 
+impl AuthZ {
+    /// Returns whether content-based authorization is active.
+    ///
+    /// Authorization is considered active if enforcement is enabled or at least one key is
+    /// configured. Without enforcement, authorization checks are still performed and reported but
+    /// failures will not result in `403 Unauthorized`
+    pub fn is_active(&self) -> bool {
+        self.enforce || !self.keys.is_empty()
```
jan-auer, line 368: This is because `enforce` defaults to `true`: We must activate auth also when there are no keys provided, so that the default config triggers auth failures.

objectstore-server/src/auth/service.rs
```diff
@@ -44,49 +44,37 @@ impl AuthAwareService {
...
+            && let Err(error) = auth.assert_authorized(perm, context)
+        {
+            sentry::with_scope(
+                |s| s.set_tag("perm", perm.to_string()),
+                || error.log(!self.enforce),
+            );
+
+            if self.enforce {
+                return Err(error.into());
+            }
```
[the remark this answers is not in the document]
jan-auer: `error.log()` creates a tracing event, which is captured as sentry event.

matt-codecov (approved):

lcian (approved):
I mainly looked at `objectstore-server/src/config.rs` and the change makes sense to me.

PR #520, opened 2026-06-25
build(deps): bump sentry to 0.48.3

lcian:
Bumps sentry to the latest version, which includes client reports, among other things.
Also removes the explicit sentry-core dependency, which is not needed.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -71,7 +71,6 @@ reqwest = { version = "0.13.4", default-features = false }
 rustls = { version = "0.23.40", default-features = false }
 secrecy = "0.10.3"
 sentry = { version = "0.48.3" }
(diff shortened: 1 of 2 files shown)


Cargo.toml
```diff
@@ -70,8 +70,7 @@ regex = "1.12.4"
 reqwest = { version = "0.13.4", default-features = false }
 rustls = { version = "0.23.40", default-features = false }
 secrecy = "0.10.3"
-sentry = { version = "0.48.2" }
-sentry-core = "0.48.2"
```
jan-auer, line 74: This was supposed to become a workspace dependency in #507, but I unintentionally reverted it. Given that we want this *just* in the client though, we can make an exception here.

jan-auer (approved):

PR #522, opened 2026-06-25
feat(server): Add on-demand heap profiling endpoints

jan-auer:
Adds gated heap profiling to production builds using jemalloc's built-in sampling. Profiling is compiled in via a new `profiling` Cargo feature (enabled in release/CI builds, off for debug) and dormant at startup. It can be activated on demand via HTTP endpoints, which dump a symbolized gzipped pprof profile.

- Outputs symbolized `pprof`, so profiles are self-contained and analyzable with `go tool pprof` without shipping the binary separately.
- Endpoints are loopback-only (peer IP check), reachable via `kubectl port-forward` or an ephemeral debug container. Normal service/ingress traffic never presents a loopback peer, so no separate listener or auth are needed.
- Profile sampling rate is set to 512 KiB mean interval which has negligible steady-state overhead without the feature being active.
- By default, sampling never runs unless explicitly enabled via the endpoint.

The CPU overhead with profiling enabled on typical workloads is about 5%. There is no measurable overhead while profiling is disabled.

Ref FS-318

### DIFF
--- a/objectstore-server/src/main.rs
+++ b/objectstore-server/src/main.rs
@@ -12,6 +12,14 @@ use objectstore_server::cli;
 #[global_allocator]
 static GLOBAL: tikv_jemallocator::Jemalloc = tikv_jemallocator::Jemalloc;
 
+// Prof enabled at compile time; sampling starts dormant and is activated on demand via the
+// /debug/pprof/enable endpoint. lg_prof_sample:19 means one sample per 2^19 = 512 KiB allocated
+// on average (jemalloc's default). The env var MALLOC_CONF still overrides this if set.
+#[cfg(all(target_os = "linux", feature = "profiling"))]
+#[allow(non_upper_case_globals)]
+#[unsafe(export_name = "malloc_conf")]
+pub static malloc_conf: &[u8] = b"prof:true,prof_active:false,lg_prof_sample:19\0";
+
 fn main() {
     match cli::execute() {
         Ok(()) => std::process::exit(0),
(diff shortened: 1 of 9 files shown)


objectstore-server/src/main.rs
```diff
@@ -12,6 +12,14 @@ use objectstore_server::cli;
 #[global_allocator]
 static GLOBAL: tikv_jemallocator::Jemalloc = tikv_jemallocator::Jemalloc;
 
+// Prof enabled at compile time; sampling starts dormant and is activated on demand via the
+// /debug/pprof/enable endpoint. lg_prof_sample:19 means one sample per 2^19 = 512 KiB allocated
+// on average (jemalloc's default). The env var MALLOC_CONF still overrides this if set.
+#[cfg(all(target_os = "linux", feature = "profiling"))]
+#[allow(non_upper_case_globals)]
+#[unsafe(export_name = "malloc_conf")]
+static malloc_conf: &[u8] = b"prof:true,prof_active:false,lg_prof_sample:19\0";
```
lcian, line 21: Do we need anything else to match the defaults we were already running with here?

lcian (approved):
Looks good to me.

objectstore-server/src/main.rs
jan-auer: This config is additive, so it doesn't remove defaults that we don't specify here. 

There are actually quite a lot of possible configs, also unrelated to profiling, that are documented [here](https://jemalloc.net/jemalloc.3.html#opt.lg_prof_sample). Even `lg_prof_sample` is set to the default, but I prefer to be explicit.

PR #526, opened 2026-06-26
fix(sentry): Bind hub to response bodies via middleware

jan-auer:
When an axum handler returns a streaming body via `Body::from_stream()`, the `IntoResponse` conversion runs inside the handler (within the sentry middleware), but the body's frames are polled by hyper **after** middleware has unwound. This means `Hub::current()` inside a stream producer resolves to whatever hub the tokio worker thread happens to have — not the request's hub.

Three endpoints are affected: `object_get`, `complete` (multipart), and `batch`.

Instead of wrapping individual streams per-endpoint, this PR introduces a middleware that wraps every response body in a `SentryBody` that re-activates the request's hub on each `poll_frame` call. This still allows for plain bodies to provide a `content-length`. It effectively double-boxes every body, but the overhead of the indirection is negligible.

Note that this does **not** cover tracing, since the Rust SDK ends the transaction when the request unwinds through middlewares. This would be addressed best by moving the SDK to span-first ingestion.

Alternative to #524.

Close FS-383


lcian (approved):
Tested in sandbox and it works.

PR #527, opened 2026-06-26
build: Disable in-process symbolication for heap profiles

jan-auer:
In-process symbolication via `addr2line`/`gimli` allocates ~400MB of memory when resolving DWARF debug info during a heap dump. This dominated heap profiles, making them useless for finding actual leaks.

Drop the `symbolize` feature from `jemalloc_pprof` so profiles contain raw addresses. Local symbolication requires a standalone installation of the latest `pprof` version.

Ref FS-318

### DIFF
--- a/README.md
+++ b/README.md
@@ -262,22 +262,34 @@ curl -X POST http://localhost:8888/debug/pprof/disable
 ```
 
 Profiles are unsymbolicated — you need the server binary to resolve stack
-frames. Copy it from the Docker image with `docker cp <image>:/bin/entrypoint
-./objectstore`.
-
-Analyze the profile dump with `go tool pprof`:
+frames. Copy it from the Docker image with:
 
 ```sh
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -261,13 +261,20 @@ curl http://localhost:8888/debug/pprof/heap > heap.pb.gz
 curl -X POST http://localhost:8888/debug/pprof/disable
 ```
 
+Profiles are unsymbolicated — you need the server binary to resolve stack
+frames. Copy it from the Docker image with `docker cp <image>:/bin/entrypoint
+./objectstore`.
```
jan-auer: We'd actually have to call this `entrypoint` and add `PPROF_BINARY_PATH=.` for pprof, but it seemingly cannot symbolize elf on macos. Investigating.

lcian (approved):

PR #528, opened 2026-06-29
fix(server): Remove service tag from request metric

jan-auer:
Investigations of memory increase over time revealed the top contributor is the
metric captured by `EmitMetricsGuard::drop`. We suspect this is the `service`
tag, which comes through the `x-downstream-service` request header and contains
the full K8S pod name. This creates a high number of aggregator buckets for the
metrics exporter's reservoir sampler.

We still have the same tag on the `server.requests` counter metric, so it is
safe to remove from the duration.

Ref FS-318

### DIFF
--- a/objectstore-server/src/web/middleware.rs
+++ b/objectstore-server/src/web/middleware.rs
@@ -111,7 +111,6 @@ pub async fn emit_request_metrics(mut request: Request, next: Next) -> Response 
 struct EmitMetricsGuard<'a> {
     route: &'a str,
     method: Method,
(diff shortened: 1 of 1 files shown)


lcian:
Should we just truncate it instead of removing?

lcian (approved):

jan-auer:
> Should we just truncate it instead of removing?

Ideally, yes. This would be best done at the source - objectstore in code has no concept of what the contents of the header are. I will look at this separately.

PR #532, opened 2026-07-01
ref(service): Use static context strings and helper functions for errors

jan-auer:
Use static string references for error context where possible, avoiding unnecessary allocations for fixed messages. Where context must be dynamic (e.g. including a key name or endpoint in the message), accept owned strings through a flexible parameter type rather than requiring callers to manually convert.

Introduce ergonomic helper functions for constructing errors so that call sites can provide context and cause directly without manually boxing or assembling struct literals.


jan-auer:
Closing in favor of #493

PR #533, opened 2026-07-01
fix(stresstest): Use `put_path` for `many` requests

lcian:
The batch workload built its requests through `put_stream()`, which always produces a `PutBody::Stream`. The client's batch classifier treats streamed bodies as unknown-size and routes them down the individual (non-batch) path, so every "batch" put in this workload was silently sent as an individual request — the batch endpoint was never actually being stress-tested.

To be classifiable as batchable, an operation needs a known body size. Rather than buffer each payload into a `Vec<u8>` in memory (which would make peak memory scale with `concurrency * batch_size * payload_size` — tens of MiB per in-flight batch at large batch sizes), each generated payload is streamed to a temp file and enqueued with `put_path()`.

### DIFF
--- a/stresstest/src/stresstest.rs
+++ b/stresstest/src/stresstest.rs
@@ -12,8 +12,8 @@ use futures::StreamExt;
 use indicatif::{ProgressBar, ProgressStyle};
 use objectstore_client::{ExpirationPolicy, Usecase};
 use sketches_ddsketch::DDSketch;
+use tokio::io::AsyncReadExt;
 use tokio::sync::Semaphore;
-use tokio_util::io::ReaderStream;
 use yansi::Paint;
 
 use crate::http::HttpRemote;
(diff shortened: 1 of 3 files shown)


stresstest/src/stresstest.rs
```diff
@@ -420,16 +419,29 @@ async fn run_batch_workload(
...
-                        many = many.push(
-                            session
-                                .put_stream(stream)
-                                .compression(None)
-                                .key(key),
-                        );
+                        let size = payload.len;
+                        payload_info.insert(key.clone(), (internal_id, size));
+                        let path = temp_dir.join(&key);
+                        let mut file = tokio::fs::File::create(&path)
```
jan-auer, line 438: Beware that on longer-running tests this could become quite a large directory. May require mounting a volume for stresstests.

jan-auer (approved):

stresstest/src/stresstest.rs
lcian: Indeed, I had to set up the `ephemeral-storage` resource on the pods, but buffering everything seems worse as it may require a ton of memory.

PR #535, opened 2026-07-02
fix(server): Ignore client-supplied read-only metadata on writes

jan-auer:
`Metadata` carries fields a client must not set on a write: `time_created` and
`time_expires` are server-computed and `size` is resolved by the backend. The write
endpoints parsed request metadata with `Metadata::from_headers`, which reads every
field, so a client could supply e.g. `x-sn-time-expires` on a write. On the local_fs
backend that forged expiration was persisted and echoed back verbatim, decoupled from
the object's actual expiration policy.

This adds `Metadata::from_insert_headers`, a parser for the untrusted write direction:
it stamps `time_created` and drops the read-only/output attributes clients cannot set,
so they can no longer be forged via headers. The create, overwrite, multipart, and
batch paths now use it, which also removes the previous per-endpoint `time_created`
stamping. `from_headers` stays the full, trusted parser for server-emitted metadata —
GET/HEAD/batch responses in the client and backend storage reads — so callers still
receive the resolved `time_expires`.

A follow-up will make this safer structurally by splitting the incoming
(client-provided) and outgoing (server-emitted) metadata into distinct types, so
read-only fields and required normalizations are enforced by the type system rather
than by hand.

### DIFF
--- a/objectstore-server/src/endpoints/batch.rs
+++ b/objectstore-server/src/endpoints/batch.rs
@@ -1,5 +1,4 @@
 use std::sync::Arc;
-use std::time::SystemTime;
 
 use axum::Router;
 use axum::extract::{DefaultBodyLimit, State};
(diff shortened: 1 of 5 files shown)


lcian (approved):
Looks good to me.

PR #536, opened 2026-07-02
fix(service): Resolve object expiration once on write

jan-auer:
Object expiration timestamps were resolved independently by each backend, which
recomputed `now() + ttl` at write time. That let the persisted `time_created` and
the effective `time_expires` drift apart, and in a tiered setup the high-volume and
long-term backends could stamp different expirations for the same object.

`time_expires` is now resolved once directly when constructing metadata from
headers (`Metadata::from_insert_headers`) — from the expiration policy and the same
`now` used for `time_created` — and every backend persists that value verbatim (GCS
`customTime`, the S3 custom-time header, and the Bigtable GC cell timestamp).
Objects without a timeout policy carry no expiration, as before. This ensures the
timestamps stay consistent across tiered storage.

TTI idle bumps are intentionally out of scope. They still recompute `now() + tti`
at the bump site and write the refreshed value into `time_expires` so the shared
write helpers stay uniform; making the bump path itself the single source of truth
is left to a follow-up (marked with TODOs).

Stacked on #535.

### DIFF
--- a/objectstore-server/docs/architecture.md
+++ b/objectstore-server/docs/architecture.md
@@ -57,24 +57,22 @@ A request flows through several layers before reaching the storage service:
 
 1. **Middleware**: metrics collection, in-flight request tracking, panic
    recovery, Sentry transaction tracing, distributed tracing.
-2. **Extractors**: path parameters are parsed into an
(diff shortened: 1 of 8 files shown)


lcian (approved):
Looks good to me.

PR #538, opened 2026-07-03
fix(gcs+s3): Drain response bodies

lcian:
Introduces a `drain_body` util on our extension trait to consume reqwests' response bodies (works on `reqwest::Response` and `Result<reqwest::Response, reqwest::Error`, and uses it where needed.
This is needed to allow `reqwest` to reuse a connection cleanly, without an explicit reset.

Without this, too many calls where the body is ignored in a short time would result in an internal error from `h2` itself: `too_many_internal_resets`. This was tested with PUT requests to GCS, where we observed around 1% of requests in the test failing due to that error.

### DIFF
--- a/objectstore-service/src/backend/common.rs
+++ b/objectstore-service/src/backend/common.rs
@@ -264,6 +264,15 @@ impl TieredWrite {
     }
 }
 
+/// Drains the body of a response we are otherwise done with.
+///
+/// reqwest only returns a connection to its pool once the response body has been fully read, so we
+/// need to explicitly drain it.
+/// Errors are swallowed, since the caller has already obtained everything it needs from the response.
+pub async fn consume_body(mut response: reqwest::Response) {
+    while let Ok(Some(_)) = response.chunk().await {}
+}
+
 /// Creates a reqwest client with required defaults.
 ///
 /// Automatic decompression is disabled because backends store pre-compressed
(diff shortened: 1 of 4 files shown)


objectstore-service/src/backend/common.rs
```diff
@@ -264,6 +264,15 @@ impl TieredWrite {
     }
 }
 
+/// Drains the body of a response we are otherwise done with.
+///
+/// reqwest only returns a connection to its pool once the response body has been fully read, so we
+/// need to explicitly drain it.
+/// Errors are swallowed, since the caller has already obtained everything it needs from the response.
+pub async fn consume_body(mut response: reqwest::Response) {
```
jan-auer: _nit_: This would be nice on `ResponseExt`, because then we could just chain it instead of assigning to an intermediate variable.

objectstore-service/src/backend/gcs.rs
```diff
@@ -704,17 +709,17 @@ impl Backend for GcsBackend {
...
-                                context: format!(
-                                    "GCS: 416 response with invalid Content-Range: {raw:?}"
-                                ),
-                                cause: None,
-                            });
-                        }
-                    }
+                    let err = match total {
+                        Some(total) => Error::RangeNotSatisfiable { total },
+                        None => Error::Generic {
```
jan-auer: while at it, let's use `Error::generic()` here.

jan-auer (commented):
Let's also add consume calls to the S3 backend.

jan-auer (approved):

PR #540, opened 2026-07-06
feat(types): Add pre-signed URL utils

lcian:
Adds `objectstore-types::presign` — low-level utilities for signing and verifying Ed25519-signed pre-signed URLs, as specified in the [design doc](https://app.notion.com/p/sentry/Capability-Pre-signed-URLs-3418b10e4b5d8076ba65c4382f077b82?v=21e8b10e4b5d80e8a4e5000cf4b021f2&source=copy_link).

## What this does

Pre-signed URLs let a client that holds an Ed25519 keypair produce a time-limited, request-bound URL without requiring the server to share a secret. The server only needs the corresponding public key to verify.

A signed request looks like:

```
GET /v1/objects/<usecase>/<scopes>/<key>
    ?os_timestamp=2026-04-20T13:37:00.00Z
    &os_duration=3600
    &os_kid=relay
    &os_sig=<signature>
```

The signature covers a **canonical form** of the request built from three newline-separated components:

```
<normalized method>\n
<path>\n
<canonical query string>
```

- **Normalized method**: uppercase HTTP method, with `HEAD` mapped to `GET`
- **Path**: the request path, included verbatim as transmitted/received on the wire
- **Canonical query string**: every query parameter except `os-sig`, with keys lowercased, sorted lexicographically by name and value, joined with `&`

This PR provides:
- `CanonicalRequest` — builds and holds the canonical string
- `CanonicalRequest::sign` — signs with a raw 32-byte Ed25519 signing key, returns a base64url-encoded signature
- `CanonicalRequest::verify` — verifies a base64url signature against a raw 32-byte verifying key
- Public query-parameter name constants (`PARAM_TIMESTAMP`, `PARAM_SIG`, `PARAM_KID`, `PARAM_DURATION`)
- `Error` type covering invalid keys, encoding failures, and verification failures

## Design decisions

**Ed25519 (asymmetric) over HMAC (symmetric).** The signer (e.g. Relay) only needs the private key; objectstore verifies with the public key and never has to share a secret. This fits the multi-tenant trust model where different callers hold different keypairs.

**Raw `[u8; 32]` key types instead of re-exporting `ed25519-dalek` types.** Keeps the public API decoupled from the specific Ed25519 crate. Callers pass plain byte arrays; conversion to library-specific key types happens internally.

**`HEAD` normalized to `GET` in the canonical form.** A single signed URL works for both `GET` and `HEAD` requests without re-signing.

**Path included verbatim (no re-encoding).** Avoids signature verification failures from differing URL encoding implementations between client and server. Based on the assumption that intermediate proxies pass the URL through with the original encoding.

**Query parameter keys lowercased before sorting.** Ensures the canonical form is case-insensitive for query parameter names, preventing verification failures from casing differences between client and server.

**No signed headers.** The canonical form does not include request headers — only the method, path, and query string are signed. This keeps the scheme simple and avoids issues with intermediaries modifying headers.

**Duplicate query parameters preserved, not deduplicated.** The canonical query sort is purely lexicographic over `key=value` strings, so duplicate keys survive and are ordered by value deterministically.

**`verify_strict` over `verify`.** Uses `ed25519-dalek`'s strict verification, which rejects edge-case malleability vectors that are technically within spec but undesirable in a security-sensitive path.

**Base64url without padding (`URL_SAFE_NO_PAD`).** Avoids `=` characters that would need percent-encoding in query strings.

## What's not in this PR

Expiry checking (`os_duration` / `os_timestamp`), key-ID resolution (`os_kid`), and server-side middleware wiring are follow-up work. This PR is the signing/verification primitive that those will build on.

Refs FS-386

### DIFF
--- a/objectstore-types/src/presign.rs
+++ b/objectstore-types/src/presign.rs
@@ -89,11 +89,17 @@ pub enum PresignError {
 /// normalized to `GET`, and [`X_OS_SIG`] is excluded from the canonical query
 /// string.
 ///
-/// `path` must be the request path as it appears in the URL (the same value the
-/// signer put on the wire and the verifier reads back, e.g. `uri.path()`); it is
-/// percent-encoded wholesale. `query` is the list of decoded query parameter
-/// key/value pairs; they are re-encoded canonically here, so ordering of the
-/// input does not matter.
+/// `path` is percent-encoded wholesale here (the `/` separators are deliberately
+/// not treated specially). This function is oblivious to the path's contents;
+/// correctness only requires that the signer and verifier pass **byte-identical**
+/// path strings. The natural choice is the raw path exactly as it appears in the
+/// request line — `uri.path()` on the server, and the equivalent from the built URL
+/// on the client. Do **not** percent-decode it first: decoding is lossy (`%2F` and a
+/// literal `/` collapse to the same byte), which would weaken the signature and risk
+/// a signer/verifier mismatch.
+///
+/// `query` is the list of decoded query parameter key/value pairs; they are
+/// re-encoded canonically here, so ordering of the input does not matter.
 ///
 /// `signed_headers` is the list of request headers named by `X-Os-Signed-Headers`
 /// (name/value pairs); names are lowercased, values are whitespace-trimmed, and
--- a/objectstore-types/src/presign.rs
+++ b/objectstore-types/src/presign.rs
@@ -227,9 +227,9 @@ mod tests {
             "/v1/objects/testing/_/key",
             Some(sample_query()),
         );
-        let signature = canonical.sign(&sk.as_bytes());
+        let signature = canonical.sign(sk.as_bytes());
 
-        assert_eq!(canonical.verify(&vk.as_bytes(), &signature), Ok(()));
+        assert_eq!(canonical.verify(vk.as_bytes(), &signature), Ok(()));
     }
 
     #[test]
@@ -242,7 +242,7 @@ mod tests {
             "/v1/objects/testing/_/key",
             Some(sample_query()),
         );
-        let signature = canonical.sign(&sk.as_bytes());
+        let signature = canonical.sign(sk.as_bytes());
 
         let tampered = CanonicalRequest::new(
             &Method::GET,
@@ -250,7 +250,7 @@ mod tests {
             Some(sample_query()),
         );
         assert_eq!(
-            tampered.verify(&vk.as_bytes(), &signature),
+            tampered.verify(vk.as_bytes(), &signature),
             Err(Error::VerificationFailed)
         );
     }
@@ -265,10 +265,10 @@ mod tests {
             "/v1/objects/testing/_/key",
             Some(sample_query()),
         );
-        let signature = canonical.sign(&sk.as_bytes());
+        let signature = canonical.sign(sk.as_bytes());
 
         assert_eq!(
-            canonical.verify(&other_vk.as_bytes(), &signature),
+            canonical.verify(other_vk.as_bytes(), &signature),
             Err(Error::VerificationFailed)
         );
     }
@@ -284,12 +284,12 @@ mod tests {
 
         // Not valid base64url.
         assert_eq!(
-            canonical.verify(&vk.as_bytes(), "not valid base64!!"),
+            canonical.verify(vk.as_bytes(), "not valid base64!!"),
             Err(Error::InvalidSignatureEncoding)
         );
         // Valid base64url but not a 64-byte signature.
         assert_eq!(
-            canonical.verify(&vk.as_bytes(), &URL_SAFE_NO_PAD.encode([0u8; 10])),
+            canonical.verify(vk.as_bytes(), &URL_SAFE_NO_PAD.encode([0u8; 10])),
             Err(Error::InvalidSignatureEncoding)
         );
     }
--- a/objectstore-types/src/presign.rs
+++ b/objectstore-types/src/presign.rs
@@ -102,7 +102,8 @@ impl CanonicalRequest {
                     let key = key.to_ascii_lowercase();
                     (key != "x-os-sig").then(|| format!("{key}={value}"))
                 } else {
-                    Some(pair.to_ascii_lowercase())
+                    let key = pair.to_ascii_lowercase();
+                    (key != "x-os-sig").then_some(key)
                 }
             })
             .collect();
@@ -206,6 +207,16 @@ mod tests {
              x-os-key-id=relay&\
              x-os-timestamp=1985-04-12T23:20:50.52Z"
         );
+    }
+
+    #[test]
+    fn key_only_sig_param_is_excluded() {
+        let canonical = CanonicalRequest::new(
+            &Method::GET,
+            "/v1/objects/testing/_/key",
+            Some("a=1&X-Os-Sig&b=2"),
+        );
+        assert_eq!(canonical.0, "GET\n/v1/objects/testing/_/key\na=1&b=2");
     }
 
     #[test]
--- a/objectstore-types/src/presign.rs
+++ b/objectstore-types/src/presign.rs
@@ -210,16 +210,6 @@ mod tests {
     }
 
     #[test]
-    fn key_only_sig_param_is_excluded() {
-        let canonical = CanonicalRequest::new(
-            &Method::GET,
-            "/v1/objects/testing/_/key",
-            Some("a=1&X-Os-Sig&b=2"),
-        );
-        assert_eq!(canonical.0, "GET\n/v1/objects/testing/_/key\na=1&b=2");
-    }
-
-    #[test]
     fn head_is_normalized_to_get() {
         let path = "/v1/objects/testing/_/key";
         assert_eq!(
--- a/objectstore-types/src/presign.rs
+++ b/objectstore-types/src/presign.rs
@@ -7,10 +7,10 @@
 //!
 //! ```text
 //! GET /v1/objects/<usecase>/<scopes>/<key>
-//!     ?X-Os-Timestamp=2026-04-20T13:37:00.00Z
-//!     &X-Os-Expires=3600
-//!     &X-Os-Key-Id=relay
-//!     &X-Os-Sig=<signature>
+//!     ?os-timestamp=2026-04-20T13:37:00.00Z
+//!     &os-duration=3600
+//!     &os-kid=relay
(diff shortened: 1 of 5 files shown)


lcian:
bugbot run
@sentry review

lcian:
bugbot run
@sentry review

objectstore-types/src/presign.rs
```diff
@@ -0,0 +1,436 @@
...
+//!   whitespace stripped, sorted by name, and joined with a `\n` separator.
+//!   When no headers are signed this component is empty, so the canonical form ends
+//!   with a trailing newline.
+
+use base64::Engine as _;
+use base64::engine::general_purpose::URL_SAFE_NO_PAD;
+use ed25519_dalek::{Signature, Signer};
+use http::{HeaderMap, HeaderName, Method};
+
+pub use ed25519_dalek::{SigningKey, VerifyingKey};
```
jan-auer: By re-exporting those types, we technically have to do a breaking change and major version bump every time `ed25519_dalek` has another semver bump. We could avoid this by wrapping the type and exposing just the functionality we need. Could you check if this is practical? 

Alternatively, we could take strings/bytes for the keys and do the conversion internally.

objectstore-types/src/presign.rs
```diff
@@ -0,0 +1,436 @@
...
+
+/// Query parameter carrying the time at which the request was signed.
+pub const X_OS_TIMESTAMP: &str = "X-Os-Timestamp";
```
jan-auer: These query params are all prefixed, but this is _our_ API. Could we strip the prefix for a cleaner API?

objectstore-types/src/presign.rs
```diff
@@ -0,0 +1,436 @@
...
+pub const X_OS_EXPIRES: &str = "X-Os-Expires";
+
+/// Query parameter carrying the list of signed headers.
+pub const X_OS_SIGNED_HEADERS: &str = "X-Os-Signed-Headers";
+
+/// Query parameter naming the key ID used to sign the request.
+pub const X_OS_KEY_ID: &str = "X-Os-Key-Id";
+
+/// Query parameter carrying the signing algorithm specifier.
+pub const X_OS_ALG: &str = "X-Os-Alg";
```
jan-auer: Putting this up for discussion: Shouldn't the algorithm follow from the key? So when the key is identified by its ID, the server should be able to infer the algorithm and use it to verify the signature.

We need to make sure that the signer can't choose too weak of an algorithm (or worse, none), and then compromise signatures intentionally or unintentionally.

objectstore-types/src/presign.rs
```diff
@@ -0,0 +1,436 @@
...
+            "GET"
+        } else {
+            method.as_str()
+        };
+
+        let mut pairs: Vec<&str> = query
+            .unwrap_or_default()
+            .split('&')
+            .filter(|pair| !pair.is_empty())
+            .filter(|pair| pair.split_once('=').map_or(*pair, |(key, _)| key) != X_OS_SIG)
```
jan-auer: This is case-sensitive, which might be surprising. For defensive programming purposes, I'd make this a `eq_ignore_ascii_case`. Also, using lowercase query params canonically would match industry standards more.

Separately, I'd add an explicit skip list here that we can extend in the future (or a function). This would make the intent clear that there are certain query params, apart from the signature, that shall be ignored.

objectstore-types/src/presign.rs
```diff
@@ -0,0 +1,436 @@
...
+            .collect();
+        pairs.sort_unstable();
+
+        let canonical_query = pairs.join("&");
+
+        let mut header_pairs: Vec<(&str, &str)> = Vec::with_capacity(signed_headers.len());
+        for name in signed_headers {
+            let value = headers
+                .get(*name)
+                .ok_or_else(|| Error::MissingSignedHeader((*name).clone()))?;
```
jan-auer: A few potential issues here:
- Many headers have defaults. This means we will fail if someone omits a "default" header if listed when it was included in the signature. 
- Many headers are case insensitive. While we trim, we would fail if there's a case difference.
- Some headers can be specified multiple times (if the header supports comma separated lists). If this is something we need to support, we should make sure that we handle this correctly in all clients. 
- Is it really a failure if the header is missing? Maybe we need to sign a URL and ensure the _absence_ of a header. In that case, it might be better to substitute a missing header with an empty string.

objectstore-types/src/presign.rs
```diff
@@ -0,0 +1,436 @@
...
+pub const X_OS_SIGNED_HEADERS: &str = "X-Os-Signed-Headers";
+
+/// Query parameter naming the key ID used to sign the request.
+pub const X_OS_KEY_ID: &str = "X-Os-Key-Id";
+
+/// Query parameter carrying the signing algorithm specifier.
+pub const X_OS_ALG: &str = "X-Os-Alg";
+
+/// Query parameter carrying the base64url-encoded signature.
+pub const X_OS_SIG: &str = "X-Os-Sig";
```
jan-auer: Could we fold all of these into a single, salted query param to implementation details from users? The current implementation creates a lot of query params that we have to read and potentially special-case.

All of the rest is contained within, e.g. separated by colons: `?signature=<exp>:<key_id>:<hash>`. Internally, the hash formed on the "canonical request" also has expiry, key_id and so on listed. I'd probably also put the header list in there, just to have it all self contained.

A disadvantage I see is supporting forward compatibility.

objectstore-types/src/presign.rs
```diff
@@ -0,0 +1,436 @@
...
+    }
+
+    /// Returns the canonical form as a string.
+    pub fn as_str(&self) -> &str {
+        &self.0
+    }
+
+    /// Signs this canonical form with Ed25519.
+    ///
+    /// Returns the base64url-encoded  signature, suitable as the value of the
```
jan-auer:
```suggestion
    /// Returns the base64url-encoded signature, suitable as the value of the
```

objectstore-types/src/presign.rs
```diff
@@ -0,0 +1,436 @@
...
+    ) -> Result<Self, Error> {
+        let canonical_method = if *method == Method::HEAD {
+            "GET"
+        } else {
+            method.as_str()
+        };
+
+        let mut pairs: Vec<&str> = query
+            .unwrap_or_default()
+            .split('&')
```
jan-auer, line 98: The manual parsing here is important to avoid decoding mismatches. Let's add a comment why we do that over a proper query parser.

objectstore-types/src/presign.rs
lcian: That would also be a viable way to do it, with the small downside that we would need to remember this detail when provisioning new keys, making sure we use different names for different algos always.

Anyways, this parameter is only included here to showcase how the scheme could be extended when/if we have a need to switch to a different algorithm. Therefore, I'll just remove it and defer this discussion on how to choose the algorithm to that time, as it's not important at the moment and we're free to choose whichever method in the future.

objectstore-types/src/presign.rs
lcian: This seems to be a more general discussion around whether we want Objectstore to be case-sensitive or not wrt query param keys.
Given that there's no existing use of query params on any of our endpoints, we're free to make the decision now, and I'm happy to go with case-insensitive. I'll change the comparison and document this somewhere.

objectstore-types/src/presign.rs
lcian: 1. If they include it in the "signed headers" then they must specify it.
2. This shouldn't be a problem since `HeaderMap` normalizes everything (including lookup keys afaik) to lowercase, anyways I'll double check.
3. True, I'll use `get_all` to take this into account.
4. Seems like this should be a separate parameter, like "forbidden headers" or something, otherwise we could conflate empty string with absent header.

objectstore-types/src/presign.rs
lcian: As long as we append to the end, it will be forwards compatible.
I think separate query params are a bit easier on the eyes for debuggability, but on the other hand it would make sense to do a single param just to shorten the URLs a bit. In that case, we need to be careful about the separator.

objectstore-types/src/presign.rs
lcian: We need to do the same for path.

lcian: Discussed: let's use whatever encoding is already there, and make sure we use the same encoded components when making the request.

objectstore-types/src/presign.rs
lcian: Discussion: find a way to solve these points, or remove the signed headers altogether.

objectstore-types/src/presign.rs
lcian: Discussion: the API is also accessible through Sentry as a proxy, so I would like to avoid any possible conflicts (like our current `Authorization` vs `X-Os-Auth` header).

objectstore-types/src/presign.rs
lcian: Removed signed headers, as they're not needed for now.

objectstore-types/src/presign.rs
lcian: Using a lowercase comparison now. There's no other query params we need to skip for now, so we'll just keep the ad-hoc skip for the signature itself.

objectstore-types/src/presign.rs
lcian: Keeping the prefixed query params.

objectstore-types/src/presign.rs
lcian: Changed to take bytes and do the conversion internally.

objectstore-types/src/presign.rs
lcian: We got rid of a few parameters, I prefer keeping them separate, so I'm moving forward with that.

jan-auer (approved):
LGTM

objectstore-types/src/presign.rs
```diff
@@ -0,0 +1,297 @@
...
+//! Example:
+//!
+//! ```text
+//! GET /v1/objects/<usecase>/<scopes>/<key>
+//!     ?os_timestamp=2026-04-20T13:37:00.00Z
+//!     &os_duration=3600
+//!     &os_kid=relay
+//!     &os_sig=<signature>
+//! ```
+//!
```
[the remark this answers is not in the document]
jan-auer: That's from the outdated PR description

PR #542, opened 2026-07-06
feat(service): Add span instrumentation to backends

jan-auer:
Backend operations now appear as real child spans in Sentry, turning the existing per-request and per-task transactions into full trace waterfalls. They were empty before because the sentry-tracing layer only converts `INFO`+ spans by default; the logging layer now forwards `DEBUG`+ spans (`TRACE` still ignored) and the backend spans moved to `debug`.

GCS/S3 HTTP calls and BigTable RPCs each get a per-attempt span covering the real request duration (method/URL/status, or action/gRPC status), and the tiered, changelog, and per-backend operations are annotated so the waterfall reads top-down. Spans carry small primitive fields (object id, range, part numbers, URLs); payloads, streams, and metadata are excluded.

Observability-only — no behavior changes and console logging is untouched.

Closes FS-412

### DIFF
--- a/objectstore-log/src/subscriber.rs
+++ b/objectstore-log/src/subscriber.rs
@@ -15,7 +15,8 @@ include!(concat!(env!("OUT_DIR"), "/constants.gen.rs"));
 ///
 /// When built with the `sentry` feature, also attaches a Sentry tracing layer if the Sentry client
 /// has been initialized. `ERROR` and `WARN` events become Sentry events; `INFO`/`DEBUG` become
-/// Sentry logs; `TRACE` is ignored.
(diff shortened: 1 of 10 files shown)


lcian (approved):
Looks really good!

The span op/name in Sentry is a bit verbose (as it always was).
This can be sometimes useful as it gives you the module where the call came from (useful to tell e.g. which backend), so I might just have to get used to it.
Image

jan-auer:
For a moment I was considering to strip the crate names and just leave the last module, but then I came back to the full module. Maybe this is something the UI could collapse in the future. 

We can for sure improve the span name, however. It currently just duplicates the function name, but there's more information we could put into the span like instrumentation in other environments does.

PR #543, opened 2026-07-07
feat(server): Support pre-signed URLs

lcian:
Wires the `objectstore_types::presign` utilities into the server so a request can authorize itself with a pre-signed URL instead of a JWT. When a request carries an `os_sig` query parameter, the extractor verifies the Ed25519 signature against the request's canonical form and produces a pre-authorized auth context. Because the signature binds the exact method, path, and query parameters, no scope check is needed — but the operation's permission is still checked against the signing key's `max_permissions`, so a restricted (e.g. read-only) key cannot pre-sign a more privileged operation such as `DELETE`.

`AuthContext` is now a three-variant enum — `Disabled` (auth inactive), `Preauthorized(key_id)` (valid pre-signed URL), or `Scoped` (verified JWT). The `Preauthorized` variant holds only the signing key's ID; `AuthAwareService` (which now holds an `Arc<PublicKeyDirectory>`) resolves that key's `max_permissions` at authorization time. Each configured PEM is now parsed into both a `DecodingKey` (JWT) and an `ed25519_dalek::VerifyingKey` (pre-sign).

Pre-signed URLs are accepted only for `GET`/`HEAD`/`DELETE`; other methods are rejected with `401 Unauthorized`. The validity window (`os_timestamp` + `os_duration`) is enforced and capped at one week. The `PresignParams` timestamp uses `SystemTime` with `humantime_serde` for deserialization, consistent with other timestamp fields in the codebase. The auth extractor uses a three-arm match on `(has_signature, method)` to dispatch between pre-signed verification, unsupported-method rejection, and JWT decoding, with shared error logging and enforce-mode handling at the bottom.

The extractor reads the path from `OriginalUri` (required — returns `InternalError` if missing) because `Router::nest` strips the `/v1` prefix that the client signs over. Query parameter names are currently case-sensitive (`os_sig`, `os_kid`, `os_timestamp`, `os_duration`) due to axum's `Query` extractor being case-sensitive. We can introduce our own case-insensitive extractor in a follow-up.

End-to-end tests in `tests/presigned.rs` cover the happy path for GET/HEAD/DELETE and rejection of tampered signatures, wrong keys, expired windows, over-one-week validity, and unsupported methods. Unit tests in `auth/context.rs` verify that `Preauthorized` honors `max_permissions` and denies an unknown/removed key.

Refs FS-386

### DIFF
--- a/objectstore-server/docs/architecture.md
+++ b/objectstore-server/docs/architecture.md
@@ -116,20 +116,26 @@ the exact request), or `Scoped` (a verified JWT, checked per operation).
 
 Instead of a JWT, a request may authorize itself with a **pre-signed URL**: a
 key holder signs a canonical form of the request with its Ed25519 key and encodes
-the signature and parameters entirely in the query string
-(`X-Os-Sig`, `X-Os-Key-Id`, `X-Os-Timestamp`, `X-Os-Expires`, and an optional
-`;`-separated `X-Os-Signed-Headers`). See [`objectstore_types::presign`] for the
+the signature and parameters entirely in the query string (`os-sig`, `os-kid`,
+`os-timestamp`, `os-duration`). See [`objectstore_types::presign`] for the
 canonical form.
 
-When the extractor sees an `X-Os-Sig` query parameter it takes the pre-signed
+Query parameter keys in the pre-signing scheme are always matched
+case-insensitively, both for these reserved parameters and for any other
+parameter folded into the canonical query string — this is a deliberate
+design choice. It does not extend to unrelated query parameters used
+elsewhere in the API (e.g. multipart upload parameters), which remain
+case-sensitive.
+
+When the extractor sees an `os-sig` query parameter it takes the pre-signed
 path instead of looking for a JWT:
 
 - Only `GET`, `HEAD`, and `DELETE` are supported. Other methods are rejected
   (currently `401`; see the `PresignUnsupportedMethod` mapping for the intended
   `501`).
 - The signature is verified against the request's canonical form using the
-  `X-Os-Key-Id` key from the [`PublicKeyDirectory`](auth::PublicKeyDirectory).
-- The validity window (`X-Os-Timestamp` + `X-Os-Expires`) is enforced, capped at
+  `os-kid` key from the [`PublicKeyDirectory`](auth::PublicKeyDirectory).
+- The validity window (`os-timestamp` + `os-duration`) is enforced, capped at
   **one week** so a URL cannot be minted to be effectively immortal.
 
 A verified pre-signed request yields an `AuthContext::Preauthorized` carrying the
--- a/objectstore-server/docs/architecture.md
+++ b/objectstore-server/docs/architecture.md
@@ -132,9 +132,11 @@ path instead of looking for a JWT:
 - The validity window (`X-Os-Timestamp` + `X-Os-Expires`) is enforced, capped at
   **one week** so a URL cannot be minted to be effectively immortal.
 
-A verified pre-signed request yields an `AuthContext::Preauthorized`, which permits
-the request as-is — the signature already binds it to the specific method, path,
-and signed headers.
+A verified pre-signed request yields an `AuthContext::Preauthorized` carrying the
+signing key's `max_permissions`. The signature already binds the request's method,
+path, and signed headers, so no scope check is needed — but the operation's
+permission is still checked against `max_permissions`, so a restricted (e.g.
+read-only) key cannot pre-sign a more privileged operation such as `DELETE`.
 
 ## Configuration
 
--- a/objectstore-server/docs/architecture.md
+++ b/objectstore-server/docs/architecture.md
@@ -123,9 +123,7 @@ canonical form.
 When the extractor sees an `os-sig` query parameter it takes the pre-signed
 path instead of looking for a JWT:
 
-- Only `GET`, `HEAD`, and `DELETE` are supported. Other methods are rejected
-  (currently `401`; see the `PresignUnsupportedMethod` mapping for the intended
-  `501`).
+- Only `GET`, `HEAD`, and `DELETE` are currently supported.
 - The signature is verified against the request's canonical form using the
   `os-kid` key from the [`PublicKeyDirectory`](auth::PublicKeyDirectory).
 - The validity window (`os-timestamp` + `os-duration`) is enforced, capped at
@@ -136,8 +134,7 @@ signing key's ID. The signature already binds the request's method, path, and
 signed headers, so no scope check is needed — but at authorization time the
 operation's permission is resolved against that key's `max_permissions`, so a
 restricted (e.g. read-only) key cannot pre-sign a more privileged operation such
-as `DELETE`. Keeping only the key ID in the context leaves the permission lookup a
-service-level concern rather than baking it in at extraction time.
+as `DELETE`.
 
 ## Configuration
 
--- a/objectstore-server/docs/architecture.md
+++ b/objectstore-server/docs/architecture.md
@@ -120,13 +120,6 @@ the signature and parameters entirely in the query string (`os-sig`, `os-kid`,
 `os-timestamp`, `os-duration`). See [`objectstore_types::presign`] for the
 canonical form.
 
-Query parameter keys in the pre-signing scheme are always matched
-case-insensitively, both for these reserved parameters and for any other
-parameter folded into the canonical query string — this is a deliberate
-design choice. It does not extend to unrelated query parameters used
-elsewhere in the API (e.g. multipart upload parameters), which remain
-case-sensitive.
-
 When the extractor sees an `os-sig` query parameter it takes the pre-signed
 path instead of looking for a JWT:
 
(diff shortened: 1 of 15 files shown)


lcian:
bugbot run
@sentry review

lcian:
bugbot run
@sentry review

objectstore-server/src/auth/context.rs
```diff
@@ -53,6 +77,22 @@ pub struct AuthContext {
...
+///
+/// [`AuthContext::assert_authorized`] can be used to check whether a request is authorized to
+/// perform certain operations on a given resource.
+#[derive(Debug, PartialEq)]
+pub enum AuthContext {
+    /// Authorization is inactive; every operation is permitted.
+    Disabled,
+
+    /// A valid signature already authorized this exact request.
+    Preauthorized(KeyId),
```
lcian: This is a consequence of the mismatch I was voicing in the design doc/discussion.
URLs are a server-level concern, while authorization and operations are a service-level concern in our architecture, so we need to defer the check of "can this key actually authorize this particular operation" to the service.
The server can still validate the query params and signature, but doesn't know about "operations".

This can be solved by implementing "pre-signed URLs" by extending our JWTs instead of doing a real signature. In that case, everything would happen at the service level.

We went back and forth on which approach to use many times, so please let's check one last time that we're all okay with this setup, or let's decide once and for all that we implement this by simply extending the JWTs.

lcian:
bugbot run
@sentry review

objectstore-server/src/auth/context.rs
```diff
@@ -127,29 +162,97 @@ impl AuthContext {
...
+        let verified = key_config.key_versions.iter().any(|key| {
+            canonical
+                .verify(key.verifying_key.as_bytes(), &params.signature)
+                .is_ok()
+        });
+        if !verified {
+            return Err(AuthError::VerificationFailure);
+        }
+
+        Ok(AuthContext::Preauthorized(params.key_id.clone()))
```
jan-auer: On first glance, it looks like we can reduce variants and create the same variant as for full tokens here if we resolve the key right away and then populate permissions from the key's max permissions. We can populate usecase and scopes from the request.

objectstore-server/docs/architecture.md
```diff
@@ -108,6 +108,34 @@ the token's scopes and permissions cover the requested
...
+Its [`AuthContext`](auth::AuthContext) is one of `Disabled` (auth inactive — all
+operations permitted), `Preauthorized` (a valid pre-signed URL already authorized
+the exact request), or `Scoped` (a verified JWT, checked per operation).
+
+### Pre-signed URLs
+
+Instead of a JWT, a request may authorize itself with a **pre-signed URL**: a
+key holder signs a canonical form of the request with its Ed25519 key and encodes
+the signature and parameters entirely in the query string (`os-sig`, `os-kid`,
+`os-timestamp`, `os-duration`). See [`objectstore_types::presign`] for the
```
jan-auer:
```suggestion
the signature and parameters entirely in the query string (`os_sig`, `os_kid`,
`os_timestamp`, `os_duration`). See [`objectstore_types::presign`] for the
```

Same below

objectstore-server/src/auth/context.rs
```diff
@@ -127,29 +162,97 @@ impl AuthContext {
...
+    pub fn from_presigned_request(
+        method: &Method,
+        path: &str,
+        raw_query: Option<&str>,
+        params: &PresignParams,
+        key_directory: &PublicKeyDirectory,
+        now: SystemTime,
+    ) -> Result<AuthContext, AuthError> {
+        let key_config = key_directory.keys.get(&params.key_id).ok_or_else(|| {
+            AuthError::InternalError(format!("Key `{}` not configured", &params.key_id))
```
jan-auer: This variant is meant for server errors rather than auth failures. Let's use or introduce an error variant that responds with 401 or 403.

objectstore-server/src/auth/context.rs
```diff
@@ -127,29 +162,97 @@ impl AuthContext {
...
+            return Err(AuthError::BadRequest(
+                "presigned URL validity exceeds the maximum of 1 week",
+            ));
+        }
+
+        let expiry = params
+            .timestamp
+            .checked_add(duration)
+            .ok_or(AuthError::VerificationFailure)?;
+        if now < params.timestamp || now > expiry {
```
jan-auer: There can be clock skew between the signing system and objectstore server. If the signing clock is just a few seconds ahead, the request will immediately be rejected.

I'd suggest to add a small grace window somewhere around 10-60 seconds.

objectstore-server/src/auth/key_directory.rs
```diff
@@ -2,17 +2,37 @@ use std::collections::{BTreeMap, HashSet};
...
 
 use crate::config::{AuthZ, AuthZVerificationKey};
 
-async fn read_key_from_file(filename: &Path) -> anyhow::Result<DecodingKey> {
+/// A single version of a public key, usable for both JWT verification (via
+/// [`DecodingKey`]) and pre-signed URL verification (via [`VerifyingKey`]).
+///
+/// Both are parsed from the same EdDSA PEM file.
+#[derive(Debug)]
+pub struct PublicKey {
```
jan-auer, line 17: This is now part of the public interface. If this is unintentional, we should make this `pub(super)` or move to an internal module.

objectstore-server/src/extractors/service.rs
```diff
@@ -22,31 +25,85 @@ impl FromRequestParts<ServiceState> for AuthAwareService {
...
+            ));
         }
 
-        let token = parts
-            .headers
-            .get(OBJECTSTORE_AUTH_HEADER)
-            .or_else(|| parts.headers.get(header::AUTHORIZATION))
-            .and_then(|v| v.to_str().ok())
-            .and_then(strip_bearer);
+        let auth_result = match has_signature(parts.uri.query()) {
```
jan-auer: We now have two branches with equal weight: signed URLs and token-based auth. To improve readability, we should reorganize the code here and introduce helper functions to emphasize the main control flow.

jan-auer (commented):
Thank you, first review pass below. The main question is on refactoring of `AuthAwareService`.

objectstore-server/src/auth/context.rs
lcian: The permission in the `AuthContext` is supposed to be the granted permission.
If the pre-signed URL is for a `GET`, it's conceptually incorrect to populate `AuthContext` with `GET+PUT+DELETE` (which would happen if those happen to be the key's `max_permissions`), as that URL only really authorizes you to do a `GET`.

We could also infer the operation from the URL and then check it agains the key's `max_permissions`, but that means that we would have to maintain a mapping of which URLs+methods map to which operations.
This is practically easy to do but conceptually I don't like it.

Please see https://github.com/getsentry/objectstore/pull/543#discussion_r3542874798.

objectstore-server/src/auth/context.rs
lcian: Introduced `AuthError::UnknownKey`, mapping to 401, and changed the JWT path to also return it.

objectstore-server/src/auth/key_directory.rs
lcian: It needs to be `pub` because `PublicKeyConfig` which uses it is also `pub`, which in turn is used by `PublicKeyDirectory`, which is used in `objectstore-server/src/state.rs`.

objectstore-server/src/extractors/service.rs
lcian: Changed.

objectstore-server/src/auth/context.rs
lcian: Good catch, added 60 secs window.

objectstore-server/src/auth/context.rs
```diff
@@ -127,29 +162,98 @@ impl AuthContext {
...
+        key_directory: &PublicKeyDirectory,
+        now: SystemTime,
+    ) -> Result<AuthContext, AuthError> {
+        let key_config = key_directory
+            .keys
+            .get(&params.key_id)
+            .ok_or(AuthError::UnknownKey)?;
+
+        let duration = Duration::from_secs(params.duration_secs);
+        if duration > MAX_PRESIGN_DURATION {
```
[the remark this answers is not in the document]
lcian: `enforce = false` means don't enforce, so I think this is the intended way.

objectstore-server/src/auth/context.rs
jan-auer: Conceptually it's a little less clean, but I don't find it too incorrect to populate `AuthContext` with a static `GET` for now - the validation of the method has to happen before anyway. If we ever implement pre-signed URLs for anything other than that, there will be more refactoring needed in both cases.

From my end, good to move forward either way, but consider if this added complexity actually provides us some benefit.

jan-auer (approved):

objectstore-server/src/auth/context.rs
lcian: Discussed on Slack.
We'll remove `DELETE` from the set of allowed methods, and move forward with just `HEAD`+`GET`, so we can just hardcode the check on `objects.get` for now.

PR #544, opened 2026-07-08
feat(py-client): Support creating pre-signed URLs

lcian:
Adds `Session.presigned_object_url("GET" | "HEAD", key, duration)` to the Python client, so a service holding an EdDSA keypair can hand out a time-limited URL that authorizes a single GET/HEAD without an auth token.

The subtle part is encoding. The server verifies the signature against the raw, still percent-encoded path and query it receives, so the client must sign the exact bytes its HTTP client puts on the wire.
urllib3 percent-encodes request targets via `urllib3.util.url._encode_target` in `HTTPConnectionPool.urlopen`, so rather than reimplement RFC 3986 and risk drift, `presign.py` reuses those same urllib3 internals — the match then holds by construction.
A unit test bench asserts parity with urllib3's wire encoding across corner cases, and end-to-end tests fetch the signed URLs against a live server with stdlib `urllib` to prove they work for external consumers.

Breaking change: `TokenGenerator` was renamed to `SecretKey` because with pre-signed URL support, the key is no longer just a JWT generator — it's a general-purpose EdDSA signing identity used for both JWT auth tokens and URL signatures. Related rename: `sign_for_scope()` → `token_for_scope()` (clarifies it returns a JWT token).
I've considered keeping `TokenGenerator` as a deprecated alias for `SecretKey`, but IMO we might as well remove it directly, as migration is trivial, and we'll be the ones doing the migration in Sentry itself.

Refs FS-386

### DIFF
--- a/clients/python/README.md
+++ b/clients/python/README.md
@@ -188,12 +188,18 @@ client = Client("http://localhost:8888", token=token)
 
 A **pre-signed URL** is a time-limited URL that authorizes a single request on
 one object without the recipient needing an auth token. This is useful for
-handing a download link to a browser or an external service.
+handing a download link to a browser or an external service, or for returning
+an HTTP redirect so that clients can download objects directly from
+Objectstore.
+
+Pre-signed URLS currently only support `GET` and `HEAD` (a pre-signed URL for
+`GET` authorizes `HEAD` too, and viceversa), so for uploads you should use
+scoped JWTs, as described above.
 
 `Session.presigned_object_url` signs the URL with the session's `SecretKey`,
-so it requires a `SecretKey` (a static token string cannot sign) and raises
-`ValueError` otherwise. Only `GET` and `HEAD` may be pre-signed, the granted
-permissions are those configured server-side for the signing key, and the
+so it requires a `SecretKey` and raises `ValueError` otherwise.
+Only `GET` and `HEAD` may be pre-signed, the granted permissions are those
+configured server-side for the signing key, and the
 validity may not exceed one week.
 
 ```python
--- a/clients/python/README.md
+++ b/clients/python/README.md
@@ -192,9 +192,9 @@ handing a download link to a browser or an external service.
 
 `Session.presigned_object_url` signs the URL with the session's `TokenGenerator`
 keypair, so it requires a `TokenGenerator` (a static token string cannot sign)
-and raises `ValueError` otherwise. Only `GET`, `HEAD`, and `DELETE` may be
-pre-signed, the granted permissions are those configured server-side for the
-signing key, and the validity may not exceed one week.
+and raises `ValueError` otherwise. Only `GET` and `HEAD` may be pre-signed,
+the granted permissions are those configured server-side for the signing key,
+and the validity may not exceed one week.
 
 ```python
 from datetime import timedelta
--- a/clients/python/README.md
+++ b/clients/python/README.md
@@ -190,7 +190,7 @@ A **pre-signed URL** is a time-limited URL that authorizes a single request on
 one object without the recipient needing an auth token. This is useful for
 handing a download link to a browser or an external service.
 
-`Session.presigned_url` signs the URL with the session's `TokenGenerator`
+`Session.presigned_object_url` signs the URL with the session's `TokenGenerator`
 keypair, so it requires a `TokenGenerator` (a static token string cannot sign)
 and raises `ValueError` otherwise. Only `GET`, `HEAD`, and `DELETE` may be
 pre-signed, the granted permissions are those configured server-side for the
@@ -200,7 +200,7 @@ signing key, and the validity may not exceed one week.
 from datetime import timedelta
 
 # The recipient can fetch this with any HTTP client, no auth header needed.
-url = session.presigned_url("GET", "my-key", duration=timedelta(hours=1))
+url = session.presigned_object_url("GET", "my-key", duration=timedelta(hours=1))
 
 import urllib.request
 with urllib.request.urlopen(url) as resp:
--- a/clients/python/README.md
+++ b/clients/python/README.md
@@ -156,30 +156,30 @@ key = resumed.complete(new_parts + existing_parts)
 If your Objectstore instance enforces authorization, you must configure authentication
 via the `token` parameter on `Client`. It accepts either:
 
-- A **`TokenGenerator`** — for internal services that have access to an EdDSA keypair.
-  The generator signs a fresh JWT for each request, scoped to the specific usecase
-  and scope being accessed.
+- A **`SecretKey`** — for internal services that have access to an EdDSA keypair.
+  The key signs a fresh JWT for each request, scoped to the specific usecase
+  and scope being accessed, and can also sign pre-signed URLs.
 - A **`str`** — a pre-signed JWT, used as-is for every request.
   Use this for external services that receive a token from another source.
 
 ```python
 from objectstore_client import Client, Usecase
-from objectstore_client.auth import TokenGenerator
+from objectstore_client.auth import SecretKey
 
 # Option 1: Internal service with a keypair
 client = Client(
     "http://localhost:8888",
-    token=TokenGenerator(kid="my-service", secret_key="<private key>"),
+    token=SecretKey(kid="my-service", secret_key="<private key>"),
 )
 
 # Option 2: External service with a pre-signed JWT
-# Use TokenGenerator.sign_for_scope() to obtain a static token from an
+# Use SecretKey.token_for_scope() to obtain a static token from an
 # internal service, then pass it to the external consumer:
 from objectstore_client.scope import Scope
 
-token = TokenGenerator(
+token = SecretKey(
     kid="my-service", secret_key="<private key>",
-).sign_for_scope("my_app", Scope(org=42, project=1337))
+).token_for_scope("my_app", Scope(org=42, project=1337))
 
 client = Client("http://localhost:8888", token=token)
 ```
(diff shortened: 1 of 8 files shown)


clients/python/src/objectstore_client/client.py
```diff
@@ -446,6 +447,78 @@ def object_url(self, key: str) -> str:
...
+        key: str,
+        duration: timedelta,
+    ) -> str:
+        """
+        Generates a pre-signed URL authorizing a single ``method`` request on the
+        object with the given ``key``, valid for ``duration``.
+
+        The returned URL carries an Ed25519 signature in its query string, so the
+        recipient can perform the request without an auth token. It can be handed
+        to any HTTP client (``urllib``, ``requests``, a browser, ...); the URL is
```
lcian:
```
        to any HTTP client (``urllib``, ``requests``, a browser, ...); the URL is
```
I'm afraid this might not hold true in non-trivial cases that require encoding, as different HTTP clients might have subtle differences in encoding, and they probably don't have a way to force the usage of a given pre-parsed path/query string.

lcian: In https://github.com/getsentry/objectstore/pull/540#discussion_r3536931508 and our offline discussion with @jan-auer we decided that we want to sign on the **already encoded** components, so I'm committing to it.

lcian: Okay I did a test generating the pre-signed URLs in Python and hitting them with `reqwest` in Rust, turns out this is more robust than I thought, seems to work in practice:

```
  Pre-signed URL Manual Test Results

  Setup: objectstore server on localhost:18889, filesystem backend, auth enforced with the test Ed25519 keypair (test_kid).

  Phase 1: Python client PUT + presign generation

  - 28/28 keys PUT successfully (no failures)
  - 28/28 presigned GET URLs generated without error

  Phase 2: Python urllib.request fetch of presigned URLs

  - 28/28 OK — every presigned URL returned status 200 with correct body

  Phase 3: Rust reqwest fetch of the same presigned URLs

  - 28/28 OK — every presigned URL returned status 200 with correct body match

  Keys tested (all passed across all clients)

  From existing e2e suite (16):
  plain-key, with space, café-unicode, emoji-😀, plus+sign, sub!$'()*+,=delims, tilde~dot.key, 100%literal-percent, looks%20encoded, nested/path/segments, quote'apostrophe, at@sign, colon:separated, ampersand&inside, semi;colon, equals=sign

  Additional tricky keys (12):
  double  space, tabs\there, newline\nkey, mixed café 😀 +&=;, cjk-日本語, arabic-مرحبا, backslash\\key, curly{brace}key, square[bracket]key, pipe|key, caret^key, backtick`key

  Conclusion

  Zero failures. The pre-signed URL signing in the Python client produces URLs that verify correctly on the server when consumed by both Python's urllib and Rust's reqwest — including all encoding corner cases (unicode, multi-byte, special delimiters,
  percent-literal, whitespace variants, CJK, Arabic).
```

lcian: Tested also with `curl`:
```
⏺ 28/28 OK with curl — all presigned URLs work perfectly. That's now three independent HTTP clients (Python urllib, Rust reqwest, curl) all successfully consuming the same presigned URLs across all 28 encoding corner cases with zero failures.
```

clients/python/tests/test_e2e.py
```diff
@@ -678,3 +680,151 @@ def put_part(part_number: int, data: bytes) -> CompletePart:
...
+ENCODING_CORNER_CASE_KEYS = [
+    "plain-key",
+    "with space",
+    "café-unicode",
+    "emoji-😀",
+    "plus+sign",
+    "sub!$'()*+,=delims",
+    "tilde~dot.key",
+    "100%literal-percent",  # `%li` isn't a valid escape, so both escape the `%`
+    # "looks%20encoded",  # diverges: quote -> %2520, urllib3 put -> %20
```
lcian, line 809: Fixed in https://github.com/getsentry/objectstore/pull/548

clients/python/src/objectstore_client/presign.py
```diff
@@ -0,0 +1,85 @@
...
+#   - sub-delims: "!$&'()*+,;="
+#   - path extras: ":" "@" "/"  (the pchar set plus the "/" segment separator)
+#   - query also allows "?"
+# This is the same set urllib3 leaves unescaped, so the encoding is a fixed
+# point of the HTTP client that ultimately sends the URL and survives on the
+# wire verbatim. It is also byte-for-byte identical to the Rust client's
+# `percent_encoding` set (see `objectstore-types/src/presign.rs`), so the two
+# clients produce the same signed bytes for any key.
+_PATH_SAFE = "/:@!$&'()*+,;="
+_QUERY_SAFE = _PATH_SAFE + "?"
```
lcian, line 26: I took a different approach initially which was using `urllib3`'s private functions (underscore-prefixed) to do the encoding.
This should be the same character set, and the encoding should be the same, the only difference is that we don't use a private API.

clients/python/src/objectstore_client/client.py
```diff
@@ -446,6 +447,67 @@ def object_url(self, key: str) -> str:
...
+
+        if not isinstance(self._token, SecretKey):
+            raise ValueError("no secret key configured on this session")
+
+        if duration > presign.MAX_PRESIGN_DURATION:
+            raise ValueError(
+                f"duration {duration} exceeds the maximum of "
+                f"{presign.MAX_PRESIGN_DURATION}"
+            )
+        duration_secs = math.ceil(duration.total_seconds())
```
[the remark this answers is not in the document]
lcian: This is okay IMO.

clients/python/src/objectstore_client/client.py
```diff
@@ -446,6 +447,67 @@ def object_url(self, key: str) -> str:
         """
         return self._make_url(key, full=True)
 
+    def presigned_object_url(
```
jan-auer, line 450: We also have the `object_url` function. Would it make sense to unify them in a way? For example:

```py
session.object_url(
    "foo/bar", 
    sign="GET",  # does not sign if not provided
    duration=timedelta(...), # optional, defaults to 1h
)
```

Just an alternative, current signature looks good to me.

clients/python/src/objectstore_client/client.py
```diff
@@ -237,13 +239,12 @@ def mint_token(
...
-        generator.
+        Raises ``ValueError`` if no ``SecretKey`` is configured
+        or if any requested permission is not granted to the key.
         """
-        if not isinstance(self._token, TokenGenerator):
-            raise ValueError("no token generator configured on this session")
-        return self._token.sign_for_scope(
+        if not isinstance(self._token, SecretKey):
+            raise ValueError("no secret key configured on this session")
+        return self._token.token_for_scope(
```
jan-auer, line 247: Can you elaborate on why we need this changed?

clients/python/src/objectstore_client/client.py
```diff
@@ -446,6 +447,67 @@ def object_url(self, key: str) -> str:
...
+        self,
+        method: Literal["GET", "HEAD"],
+        key: str,
+        duration: timedelta,
+    ) -> str:
+        """
+        Generates a pre-signed URL authorizing a single ``method`` request on the
+        object with the given ``key``, valid for ``duration``.
+
+        The returned URL carries an Ed25519 signature in its query string, so the
```
jan-auer: Let's be less concrete on what the signature is or has. it should be a opaque to the user so we can change out the implementation details at any point.

clients/python/src/objectstore_client/client.py
```diff
@@ -446,6 +447,67 @@ def object_url(self, key: str) -> str:
...
+        recipient can perform the request without an auth token. It can be handed
+        to any HTTP client (``urllib``, ``requests``, a browser, ...); the URL is
+        already percent-encoded and must be transmitted verbatim.
+
+        Only ``GET`` and ``HEAD`` may be pre-signed. The permissions granted
+        are those configured server-side for the signing key. ``duration``
+        must not exceed one week.
+
+        Raises ``ValueError`` if no ``SecretKey`` is configured on this
```
jan-auer: _nit_: It might be useful to pull this information up, as this is a quite important restriction and precondition, not just a "by the way, this error can happen".

clients/python/README.md
```diff
@@ -156,34 +156,57 @@ key = resumed.complete(new_parts + existing_parts)
...
-token = TokenGenerator(
+token = SecretKey(
     kid="my-service", secret_key="<private key>",
-).sign_for_scope("my_app", Scope(org=42, project=1337))
+).token_for_scope("my_app", Scope(org=42, project=1337))
 
 client = Client("http://localhost:8888", token=token)
 ```
 
+### Pre-signed URLs
```
jan-auer, line 187: It would be useful to talk about redirects as a possible use case for this. And also call out that for external uploads, it's better to pass a scoped token and use the full client.

jan-auer (approved):

clients/python/src/objectstore_client/client.py
lcian: It was called `sign_for_scope` but in reality it's encrypting a JWT.
`sign_for_scope` could be confused with the new pre-signed URL, so I wanted a clear distinction.

lcian: Similar reasoning for `TokenGenerator`.
It's actually a secret key that's not necessarily used to generate a token (JWT), but can also be used to sign for pre-signed URLs.

clients/python/src/objectstore_client/client.py
lcian: I'd keep it like this for now.
But we should probably have a default on the duration, we currently don't, let me add that.

clients/python/README.md
lcian: @jan-auer Why is it that for external uploads it's better to pass a scoped token?

clients/python/src/objectstore_client/client.py
lcian: I've kept the legacy aliases for now, so that this is not a breaking change.

clients/python/src/objectstore_client/presign.py
```diff
@@ -0,0 +1,85 @@
...
+SUPPORTED_METHODS = ("GET", "HEAD")
+
+
+def encode_path(path: str) -> str:
+    """Percent-encodes a request path as it will appear on the wire.
+
+    Leaves the RFC 3986 unreserved set, sub-delims, and ``:`` ``@`` ``/``
+    unescaped (see :data:`_PATH_SAFE`); everything else is percent-encoded.
+    """
+    return quote(path, safe=_PATH_SAFE)
```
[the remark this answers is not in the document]
lcian: Fixed in the follow-up PR.

PR #545, opened 2026-07-08
feat(rust-client): Mint scoped tokens through a TokenRequest builder

matt-codecov:
The Rust client can now mint tokens with narrower permissions or a different
expiry than the signing generator, matching the Python client's behavior
(#432). Token minting moves behind a `TokenRequest` builder, so callers pick
overrides before signing:

```rust
// Generator: default permissions and expiry
let token = generator.create_token(&scope).sign()?;

// Session: hand a read-only, short-lived token to an external service
let token = session
    .create_token()?
    .expect("generator is configured")
    .permissions(&[Permission::ObjectRead])
    .expiry_seconds(30)
    .sign()?;
```

Overrides can only narrow: requesting a permission the generator was not
granted fails with `Error::PermissionEscalation` listing the offending
permissions, rather than silently minting a weaker or stronger token. A
session backed by a static pre-signed token has nothing to re-sign, so
`create_token` returns `Error::StaticTokenOverride`; `get_token` still returns
that token verbatim.

Closes FS-428

### DIFF
--- a/clients/rust/src/client.rs
+++ b/clients/rust/src/client.rs
@@ -10,7 +10,7 @@ use reqwest::RequestBuilder;
 use url::Url;
 
 use crate::IntoTokenProvider;
-use crate::auth::TokenProvider;
+use crate::auth::{Permission, TokenProvider};
 
 const USER_AGENT: &str = concat!("objectstore-client/", env!("CARGO_PKG_VERSION"));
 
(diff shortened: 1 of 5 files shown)


clients/rust/src/client.rs
```diff
@@ -456,13 +456,43 @@ impl Session {
...
+    /// When `permissions` is `Some`, they must be a subset of the permissions granted to the
+    /// underlying [`TokenGenerator`](crate::TokenGenerator), otherwise an
+    /// [`Error::PermissionEscalation`](crate::Error::PermissionEscalation) is returned. When
+    /// `expiry_seconds` is `Some`, it overrides the generator's default expiry.
+    ///
+    /// These overrides only apply to a [`TokenProvider::Generator`](crate::TokenProvider). A
+    /// static, pre-signed token cannot be re-scoped, so passing an override alongside one returns
+    /// [`Error::StaticTokenOverride`](crate::Error::StaticTokenOverride). Returns `None` when no
+    /// token or generator was provided.
+    pub fn mint_token_with(
```
lcian: I think we should change both this method and `sign_with` to use the builder pattern.
We can keep `mint_token` and `sign` for backwards compatibility.
About errors, we can use the existing pattern we have on other builders in this very crate, that is, store a `Result` internally in the builder, and only return the result from `build()`. This way the caller needs to handle the error only at the `build()` call.

clients/rust/src/client.rs
```diff
@@ -456,13 +456,42 @@ impl Session {
         url
     }
 
-    /// Returns a signed token if a token or generator was provided or `None` otherwise
+    /// Returns a signed token if a token or generator was provided or `None` otherwise.
```
lcian: It can also return the static token.

clients/rust/src/client.rs
```diff
@@ -456,13 +456,42 @@ impl Session {
...
+    ///
+    /// Use this to produce a static token that can be handed to an external service
+    /// with lower permissions than the current session.
+    ///
+    /// # Errors
+    ///
+    /// This requires a [`TokenGenerator`](crate::TokenGenerator) to sign a new token.
+    /// If not available, this will result in an error. To retrieve a token with the current
+    /// session's default permissions and expiry, use [`get_token`](Self::get_token) instead.
+    pub fn create_token(&self) -> crate::Result<Option<TokenRequest<'_>>> {
```
lcian, line 500: This could be called `mint_token`. I guess you want to avoid the breaking change?

clients/rust/src/client.rs
jan-auer: That's what the doc string says:

> if a **token or generator** was provided

Let me improve that doc comment though.

clients/rust/src/client.rs
jan-auer: Correct, hence needed a new name. Sadly, the name is now completely different from Python.

lcian (approved):
Looks good to me.

clients/rust/src/client.rs
lcian: I would probably do a breaking change, however there's also no real need for it.
I guess we can just do all of them together in the next breaking release (e.g. when we remove multipart).

PR #546, opened 2026-07-08
feat(py-client): Implement "many" api for batch requests

matt-codecov:
https://github.com/getsentry/objectstore/pull/277 / https://github.com/getsentry/objectstore/pull/478 implemented the `many` API in the Rust client which uses Objectstore's batch endpoint. This PR ports it over to Python.

Incorporates direction from #419.

Closes FS-330

Some notes:
- Hand-rolls the max concurrency limit rather than rely on `ThreadPoolExecutor`'s `max_workers` to allow for results to be streamed from individual batch requests without having to buffer the whole response. Otherwise a batch of 1000 5MB GET results would eat 5GB RAM and delay yielding _anything_ to the caller.
- Kind of aggressive about raising errors. Being flexible is the server's job, not the client's.
- There's duplicated code. The `Get` / `Put` operation type classes copy the arg list of the `get()` and `put()` methods on `session`, the decompression code in `get()` is copied... but I didn't want to touch existing code much to reorganize in this PR.
- Default concurrency is `1` because that's the default urllib3 connection pool size. You can still send concurrent requests with a connection pool size of 1, it just opens/closes a connection per request and logs a warning about it instead of actually pooling.
- Actually reads the "part number" header from the batch endpoint response. Each operation's response is tagged with the operation's index in the input list so you can figure out which keyless PUT was assigned which key.
- Robot generated the tests, haven't reviewed them yet.

### DIFF
--- a/clients/python/README.md
+++ b/clients/python/README.md
@@ -151,6 +151,54 @@ existing_parts = resumed.list_parts()
 key = resumed.complete(new_parts + existing_parts)
 ```
 
+### Many API (batch operations)
+
+The Many API lets you enqueue multiple operations that the client executes using
+Objectstore's batch endpoint, minimizing network overhead. Operations too large
+to batch (oversized inserts) automatically fall back to individual requests, and
+batch and individual requests run concurrently.
+
+`send()` returns an `OperationResults` list. Results are **not** guaranteed to be
+in the order they were enqueued, since operations execute concurrently. Each
+result is a `GetResult`, `PutResult`, `DeleteResult`, `HeadResult`, or (for errors
+that can't be attributed to a specific operation) an `ErrorResult`.
+
+```python
+from objectstore_client import Client, Usecase
+from objectstore_client.many import GetResult, PutResult
+
+client = Client("http://localhost:8888")
+session = client.session(Usecase("attachments"), org=42, project=1337)
+
+results = (
+    session.many()
+    .put(b"file1 contents", key="file1")
+    .put(b"file2 contents", key="file2")
+    .get("file3")
+    .delete("file4")
+    .send()
+)
+
+for result in results:
+    if result.error is not None:
+        ...  # handle per-operation error
+    elif isinstance(result, GetResult):
+        payload = result.response.payload if result.response else None  # None => 404
+    elif isinstance(result, PutResult):
+        stored_key = result.key
+```
+
+If you don't need to inspect individual results and just want to raise if any
+operation failed, use `raise_for_failures()`, which raises an `ExceptionGroup` of
+all per-operation errors:
+
+```python
+session.many().put(b"file1", key="file1").put(b"file2", key="file2").send().raise_for_failures()
+```
+
+Concurrency is configurable via `.max_individual_concurrency(n)` (default 5) and
+`.max_batch_concurrency(n)` (default 3).
+
 ### Authentication
 
 If your Objectstore instance enforces authorization, you must configure authentication
--- a/clients/python/README.md
+++ b/clients/python/README.md
@@ -155,6 +155,127 @@ existing_parts = resumed.list_parts()
 key = resumed.complete(new_parts + existing_parts)
 ```
 
+### Many API (batch operations)
+
+`session.many()` executes a number of operations with as few requests as
+possible. Those within the batch protocol's per-part limit of 1 MB are grouped
+into requests to Objectstore's batch endpoint which cuts network overhead
+considerably. Inserts too large for that, or of unknown size, are sent as
+individual requests instead.
+
+Pass any iterable of `Get`, `Put`, `Delete`, and `Head` operations, which live in
+the `many` module. Results come back as `GetResult`, `PutResult`, `DeleteResult`,
+and `HeadResult` objects, each carrying the object's `key` and an `error` that is
+`None` when the operation succeeded:
+
+```python
+from objectstore_client import Client, Usecase, many
+
+client = Client("http://localhost:8888")
+session = client.session(Usecase("attachments"), org=42, project=1337)
+
+results = session.many(
+    [
+        many.Put(b"file1 contents", key="file1"),
+        many.Put(b"file2 contents", key="file2"),
+        many.Get("file3"),
+        many.Delete("file4"),
+        many.Head("file5"),
+    ]
+)
+
+for result in results:
+    if result.error is not None:
+        ...  # this operation failed
+    elif isinstance(result, many.GetResult):
+        # `response` is None if the object does not exist.
+        payload = result.response.payload if result.response else None
+```
+
+`session.many()` returns an `OperationResults` object which is a lazy iterator.
+As the iterator is consumed, it assembles batch requests and sends them to
+Objectstore. As responses come in, the operation results are yielded. Abandoning
+the iterator without fully consuming it will cancel whatever has not been
+dispatched yet.
+
+If successful results don't need to be processed or inspected, callers can call
+`raise_for_failures()` to drain the results and raise an `ExceptionGroup` with
+all per-operation errors, or `failures()` which returns the failed results as a
+list:
+
+```python
+session.many([many.Delete("file1"), many.Delete("file2")]).raise_for_failures()
+
+for failure in session.many([many.Delete("file3")]).failures():
+    print(failure.key, failure.error)
+```
+
+#### Concurrency
+
+`concurrency` caps how many requests are in flight, and defaults to `3`.
+Requests run on a thread pool created by `session.many()`, which is shut down
+when the results are exhausted or abandoned. When `concurrency` is set to `1`,
+requests are run serially on the caller thread instead, with no thread pool.
+
+```python
+client = Client("http://localhost:8888")
+session = client.session(Usecase("attachments"), org=42, project=1337)
+
+for result in session.many(operations, concurrency=8):
+    ...
+```
+
+Note: when a `Client` is built with custom `connection_kwargs` that include
+`"block": True`, the `concurrency` argument is clamped to the connection pool's
+configured size. An illustrative example:
+
+```python
+# Client created with a pool size of 4 and block=True
+client = Client(
+    "http://localhost:8888",
+    connection_kwargs={"maxsize": 4, "block": True},
+)
+session = client.session(Usecase("attachments"), org=42, project=1337)
+
+# `concurrency` is clamped to `4` because `block=True` was set
+for failure in session.many(operations, concurrency=16).failures():
+    print(failure.key, failure.error)
+```
+
+Results are yielded as responses are received, and the order isn't necessarily
+the same order that operations were given in. Each result carries an `index`
+field that corresponds to the index of the `Get` / `Put` / `Delete` / `Head`
+operation in the operation iterable passed into `session.many()`. This `index`
+allows a keyless `Put` operation to be linked with its result to learn the key
+that was assigned.
+
+```python
+uploads = [many.Put(b"first"), many.Put(b"second")]
+
+for result in session.many(uploads):
+    print(f"{uploads[result.index].contents!r} was stored as {result.key}")
+```
+
+A response part that cannot be attributed to any operation at all is reported
+as an `ErrorResult`, whose `index` is always `None`. Every other result carries
+the index of the operation it belongs to.
+
+Within a single batch, the Objectstore server processes individual operations
+concurrently and each operation's relative order is undefined. Two operations on
+the same key therefore race, and `session.many()` does nothing to prevent that.
+
+#### Metrics
+
+When a metrics backend is configured, `session.many()` emits some metrics:
+- `storage.batch.latency`: a timer recording a batch request's execution time,
+  tagged with a (bucketed) number of operations included in the batch
+- `stoarge.batch.operations`: a simple counter of individual operations, tagged
+  with each operation's kind (i.e. `PUT`/`GET`/`DELETE`).
+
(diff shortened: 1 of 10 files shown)


jan-auer:
Please see also the now closed #419. It's gone out of sync with main, so it's better to carry on here. 

The prior PR focused on streaming as much as possible and had gone through a round of feedback with the Python SDK maintainers for the public API. Also, there's configuration for concurrency that uses an optional thread pool executor. I hope there are some parts we could leverage from that.

clients/python/README.md
```diff
@@ -155,6 +155,121 @@ existing_parts = resumed.list_parts()
...
+```
+
+An `ErrorResult` carries `index=None` when the response part it came from could
+not be attributed to any operation at all.
+
+Within a single batch, the Objectstore server processes individual operations
+concurrently and each operation's relative order is undefined. To minimize the
+likelihood of racing operations on the same key, the client will separate
+same-key operations into different batches if at least one of the operations is
+a write or delete. However, with a `concurrency` value larger than `1`, it is
```
lcian: I don't think there's a need to separate operations on the same key into different batches, as the result will be non-deterministic anyways when concurrency > 1.

clients/python/src/objectstore_client/client.py
```diff
@@ -288,6 +294,73 @@ def _make_url(self, key: str | None, full: bool = False) -> str:
...
+                :class:`~objectstore_client.many.Delete`, and
+                :class:`~objectstore_client.many.Head` instances. Any iterable
+                works.
+            concurrency: The maximum number of requests in flight. Defaults to
+                ``1``, which runs everything sequentially on the calling thread,
+                without a thread pool. Raising it is sufficient to send requests
+                concurrently, but each request opens/closes its own connection
+                without additional configuration on the :class:`Client` (a
+                ``maxsize`` key in the ``connection_kwargs`` dict to control
+                ``urllib3`` connection pool size).
```
lcian: In the Rust client, by default, we send the requests with some degree of concurrency.
I would probably do that here too (so, default to the internal threadpool and `concurrency > 1`), probably with the same defaults as Rust.
Unless you have strong reasons to avoid that, it seems better UX to just handle this internally in the client rather than having the user necessarily think about this.

clients/python/src/objectstore_client/many.py
```diff
@@ -0,0 +1,971 @@
+"""
+Batch operations API for executing multiple get/put/delete operations.
```
lcian:
```suggestion
Batch operations API for executing multiple operations.
```
This should/will support also HEAD and whatever other ops we come up with in the future

clients/python/src/objectstore_client/many.py
```diff
@@ -0,0 +1,971 @@
...
+# Maximum (post-compression, estimated) size of a single part's body in a batch
+# request. Inserts above this are sent as individual requests instead.
+MAX_BATCH_PART_SIZE = 1024 * 1024  # 1 MB
+
+# Maximum total (post-compression, estimated) body size of a single batch request.
+MAX_BATCH_BODY_SIZE = 100 * 1024 * 1024  # 100 MB
+
+# Default maximum number of concurrent requests. Requests beyond the connection
+# pool's `maxsize` still go out (urllib3 opens a connection on demand and caps
+# only how many it keeps idle) but every connection over that limit is closed
```
lcian: >  urllib3 opens a connection on demand and caps only how many it keeps idle

This is true unless you configure the pool with `block=True` (https://urllib3.readthedocs.io/en/stable/advanced-usage.html).
Maybe we should avoid going so much into such details in these comments, I think what we have in the README and docstring on the API is enough as it also talks about this.

clients/python/src/objectstore_client/many.py
```diff
@@ -0,0 +1,971 @@
...
+        return None
+
+
+def _zstd_compress_bound(size: int) -> int:
+    """
+    Returns the worst-case size of ``size`` bytes compressed as a single zstd frame.
+
+    A port of the ``ZSTD_COMPRESSBOUND`` macro definition in ``zstd.h``.
+    """
+    margin = ((128 << 10) - size) >> 11 if size < (128 << 10) else 0
```
lcian, line 343: I assume this is correct.

clients/python/src/objectstore_client/many.py
```diff
@@ -0,0 +1,971 @@
...
+                yield _Batch(batch)
+                batch, batch_size, keys = [], 0, {}
+            # Now yield this individual op
+            yield _Individual(index, op)
+            continue
+
+        # If this op `_conflicts()` with an op in the current batch, or we've
+        # hit a batch size/length limit, cut the batch off here and yield it.
+        if batch and (
+            _conflicts(op, keys)
```
lcian: As I said previously, I would just avoid this logic and complexity, if the user requests conflicting operations it's their fault, our contract is clear on the fact that the ops can be executed in arbitrary order.

clients/python/src/objectstore_client/many.py
```diff
@@ -0,0 +1,971 @@
...
+) -> Generator[OperationResult, None, None]:
+    work = _iter_work(
+        (index, op, _classify(session, op)) for index, op in enumerate(operations)
+    )
+
+    if concurrency == 1 and executor is None:
+        for item in work:
+            yield from _run_work(session, item)
+    else:
+        yield from _execute_concurrent(session, work, concurrency, executor)
```
lcian: Do we really need the special case for concurrency=1 or could we just send the ops into `_execute_concurrent` with a special/single-threaded executor and that's it?

clients/python/src/objectstore_client/many.py
```diff
@@ -0,0 +1,971 @@
...
+                continue
+        return False
+
+    def run(item: _Work, scopes: tuple[Scope, Scope]) -> None:
+        """Work item runner that runs in the executor."""
+        isolation_scope, current_scope = scopes
+        try:
+            # Trace headers come from the ambient scope, which is thread-local.
+            # Restoring the caller's scope keeps requests in the same trace.
+            with use_isolation_scope(isolation_scope), use_scope(current_scope):
```
lcian: iirc the Sentry SDK should do this automatically via the `ThreadingIntegration`.
But, in theory, the user could be running without this integration enabled or an old version of the SDK that doesn't propagate the scopes correctly to the executor, so I guess we should keep it?

clients/python/src/objectstore_client/many.py
```diff
@@ -0,0 +1,971 @@
...
+                try:
+                    for result in work_results:
+                        if not put(result):
+                            return
+                finally:
+                    # Releases the connection if the consumer abandoned us
+                    # part-way through the batch response.
+                    work_results.close()
+        except Exception as error:
+            put(ErrorResult(None, error))
```
lcian, line 535: My clanker says:
```
  - [P2] Preserve known indexes for malformed status headers — /Users/lorenzotmp/repos/objectstore/clients/python/src/objectstore_client/many.py:876-880
    When a response part has a valid operation-index header but a missing or malformed status header, this returns position=None and ErrorResult(index=None) even though
    the operation is known. The position is therefore not added to seen, and _unanswered later emits a second failure for the same operation, violating the one-result-per-
    operation contract and discarding useful failure context.
```
This seems legit. If you get a 500 for an intermediate proxy for instance, this would happen.
We can probably avoid this by doing a better check on the response headers.

clients/python/src/objectstore_client/many.py
```diff
@@ -0,0 +1,971 @@
...
+    )
+
+    # The server sends each part as its operation completes, so a batch's latency
+    # is that of the whole exchange — including time the caller spends taking
+    # results off this generator, which is what paces reading the response.
+
+    # Measures batch processing latency. Note that this includes the time the
+    # caller spends consuming the results iterator and not just client/server
+    # latency.
+    with measure_storage_operation(
```
lcian, line 664: This is a bit weird for a user because a user doesn't think in terms of "batch" calls but rather in terms of "many" calls and/or individual ops.
But I also don't have any better approach to suggest here.

clients/python/src/objectstore_client/many.py
```diff
@@ -0,0 +1,971 @@
...
+    )
+    results: queue.Queue[OperationResult | _Done] = queue.Queue(maxsize=concurrency * 2)
+    cancelled = threading.Event()
+
+    def put(item: OperationResult | _Done) -> bool:
+        """
+        Passes a result through the results queue out to the caller. Returns
+        ``False`` if we were told to exit.
+        """
+        while not (cancelled.is_set() or _SHUTTING_DOWN.is_set()):
```
lcian, line 509: Is this the only place where we need to check `_SHUTTING_DOWN`?
(I would be happy if we could avoid that altogether but apparently not using it could cause the process to hang)

clients/python/src/objectstore_client/formdata.py
lcian, line 1: I haven't reviewed this, unfortunate that we have to hand-roll it.

clients/python/README.md
matt-codecov: this idea was taken from #419, except that PR only applied it when `concurrency=1`. this PR applies it for all values of `concurrency`. if the splitting logic is written anyway, why not apply it on all paths to at least make the race less likely?

as for whether it's necessary: tbh i'm ambivalent. if i were using the objectstore client, i'd probably take care to keep only the last PUT/DELETE operation (last write wins) and drop any post-write GET/HEAD operations because we already know what they'll return (unless clients _expect_ an out-of-band racing write operation lol). but there is a case one may want to express with a single `.many()` today that would need this fix: a `GET /foo/bar` to read the current value before a `PUT /foo/bar` to replace the value. a non-atomic swap.

this fix and `concurrency=1` are needed for that case to work right with a single `session.many()` call. but as a workaround you can do a `session.many()` with all of your `GET`s and then a second `session.many()` with all the `PUT`s. that eliminates any racing (as far as this client goes) and doesn't even require `concurrency=1`. it might be better to do it that way anyway.

so i think i've talked myself into agreeing with you. @jan-auer is there anything else to consider about this or shall i remove this behavior?

clients/python/src/objectstore_client/many.py
matt-codecov: scopes are inherited at time of creation so i think the integration would work when we create the executor, just not necessarily a passed-in executor. maybe we don't need to support a passed-in executor; the robot just said i should in a review pass.

clients/python/src/objectstore_client/formdata.py
matt-codecov: the algorithms here were taken mostly unmodified from @jan-auer's #419

clients/python/src/objectstore_client/many.py
matt-codecov: https://github.com/facebook/zstd/blob/82d322c4973d9e2968d94047a40892bc6d9a9bdf/lib/zstd.h#L249 is the source. it is a bit of a mess to parse but here it is with some extra indentation for clarity:
```c++
#define ZSTD_COMPRESSBOUND(srcSize) \
  ( \
    ((size_t)(srcSize) >= ZSTD_MAX_INPUT_SIZE) \
      ? 0 \
      : (srcSize) + ((srcSize)>>8) + (         /* size + (size >> 8) +             */ \
        ((srcSize) < (128<<10))                /* if size < (128 << 10)            */ \
          ? (((128<<10) - (srcSize)) >> 11)    /* then ((128 << 10) - size) >> 11  */ \
          : 0                                  /* else 0                           */ \
      ) \
  )
```

once upon a time writing convoluted preprocessor macros was my job

clients/python/src/objectstore_client/many.py
lcian: Hmm yeah you're right. I think it's good to support the custom executor, even though we probably won't use it right now in `sentry`. Then let's leave this as is.

clients/python/src/objectstore_client/many.py
matt-codecov: the special case here skips any executor/threading and just runs it on the caller thread. i don't think there is a "same thread executor" ready-made we could use here. but it probably wouldn't be hard to write

clients/python/src/objectstore_client/many.py
```diff
@@ -0,0 +1,965 @@
...
+    The zstd compressor is somewhat heavy so it is dropped as soon as the source
+    stream is exhausted.
+    """
+
+    def __init__(self, source: IO[bytes]):
+        compressor = zstandard.ZstdCompressor()
+        self._reader: ZstdCompressionReader | None = compressor.stream_reader(
+            source,
+            closefd=False,  # Not our stream to close
+        )
```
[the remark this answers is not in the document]
matt-codecov: `stream_reader` only creates the context internally on first read so this actually works out. however, that's a very obscure fact to rely on. so now the whole stream_reader is lazily created on first read instead.

clients/python/src/objectstore_client/many.py
```diff
@@ -0,0 +1,995 @@
...
+                pool.submit(run, item)
+                submitted += 1
+
+            # Normal completion condition has triggered
+            if exhausted and submitted == finished:
+                return
+
+            # Pop a result off the queue. If it's `_Done` then the whole work
+            # item has completed. Otherwise it's an individual operation result.
+            result = results.get()
```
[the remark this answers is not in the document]
matt-codecov: okay

clients/python/src/objectstore_client/many.py
matt-codecov: i left this as-is

PR #552, opened 2026-07-13
ref(server): Add report_only flag to bandwidth limits

jan-auer:
Adds a `report_only` flag to `BandwidthLimits`. When enabled, all byte accounting, EWMA estimation, gauge metrics, and KEDA metrics remain active, but requests that exceed the limit are admitted rather than rejected with 429. This allows disabling enforcement while keeping the metrics available for the autoscaler and general observability. Throughput limits are unaffected.

The flag defaults to `false`.

Ref FS-436


lcian (approved):
Looks good to me.

PR #553, opened 2026-07-14
ref(server): Emit bandwidth counter to Datadog

jan-auer:
Bandwidth bytes were captured as a gauge derived from the internal EWMA estimator. They are now additionally emitted as a statsd counter (`server.bandwidth.bytes`), so the rate can be computed by the consumer via `as_rate()`. This removes the dependency on the in-process EWMA for bandwidth observability — once rate limiting no longer needs the estimator, it can be removed entirely.


lcian (approved):
Nice!

PR #554, opened 2026-07-14
ref(server): Replace bandwidth EWMA with lock-free GCRA buckets

jan-auer:
Bandwidth rate limiting now uses debt-based GCRA (Generic Cell Rate Algorithm) buckets instead of EWMA estimation. Each bucket tracks a single atomic timestamp — the point at which all recorded traffic is "paid off." Admission checks compare this deadline against wall-clock time plus a configurable burst tolerance, with no background estimation task required.

This removes the 50ms EWMA tick and its associated state (per-scope/per-usecase shadow EWMA maps, the `EwmaEstimator` struct), replacing them with lock-free `AtomicU64::update` operations that are both simpler and more responsive — debt is visible immediately after bytes are recorded rather than after the next estimation tick. The `burst_ms` config parameter (default: 1000ms) controls how much transient overshoot is tolerated before rejection.

EWMA gauges are removed from the KEDA endpoint since the monotonic counters with `irate()` are strictly more responsive. The throughput background task is simplified to emit only the limit and map-size gauges at 1s intervals. The bandwidth limit gauge is now emitted once at construction.

Ref FS-436

### DIFF
--- a/objectstore-server/src/rate_limits.rs
+++ b/objectstore-server/src/rate_limits.rs
@@ -320,6 +320,8 @@ impl RateLimiter {
 #[derive(Debug)]
 struct BandwidthBucket {
     /// Nanoseconds since `epoch` at which all recorded traffic is paid off.
+    ///
+    /// Saturates at `u64::MAX` after ~584 years of uptime.
     tat: AtomicU64,
     /// Nanoseconds of TAT advance per byte: `1e9 / bps`.
     nanos_per_byte: f64,
@@ -346,13 +348,13 @@ impl BandwidthBucket {
         // (no credit accumulation), then advance by the byte cost.
         self.tat
             .update(Ordering::Relaxed, Ordering::Relaxed, |old| {
-                old.max(now_nanos) + weight
+                old.max(now_nanos).saturating_add(weight)
             });
     }
 
     /// Returns `true` if the bucket admits new traffic at `now_nanos`.
     fn check(&self, now_nanos: u64) -> bool {
-        self.tat.load(Ordering::Relaxed) <= now_nanos + self.burst_ns
+        self.tat.load(Ordering::Relaxed) <= now_nanos.saturating_add(self.burst_ns)
     }
 }
 
--- a/objectstore-server/src/rate_limits.rs
+++ b/objectstore-server/src/rate_limits.rs
@@ -472,6 +472,13 @@ impl BandwidthRateLimiter {
             buckets.push(Arc::clone(bucket));
         }
 
+        objectstore_metrics::gauge!(
+            "server.rate_limiter.bandwidth.usecase_map_size" = self.usecases.len()
+        );
+        objectstore_metrics::gauge!(
+            "server.rate_limiter.bandwidth.scope_map_size" = self.scopes.len()
+        );
+
         BandwidthHandle {
             buckets,
             total_bytes: Arc::clone(&self.total_bytes),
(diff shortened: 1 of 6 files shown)


objectstore-server/src/rate_limits.rs
```diff
@@ -361,147 +371,75 @@ fn pct_of_u64(value: u64, pct: u8) -> u64 {
...
 }
 
 impl BandwidthRateLimiter {
     fn new(config: BandwidthLimits) -> Self {
+        let global = config
+            .global_bps
+            .map(|bps| Arc::new(BandwidthBucket::new(bps, config.burst_ms)));
+
+        if let Some(limit) = config.global_bps {
+            objectstore_metrics::gauge!("server.bandwidth.limit" = limit);
```
jan-auer, line 395: This is tracked only once, but metrics-rs keeps re-sending the gauge for the lifetime of the process.

objectstore-server/src/rate_limits.rs
```diff
@@ -252,97 +274,85 @@ impl RateLimiter {
...
-            .store(ewma.floor() as u64, std::sync::atomic::Ordering::Relaxed);
+    /// Records `bytes` of consumption unconditionally (debt is allowed).
+    fn spend(&self, now_nanos: u64, bytes: u64) {
+        let weight = (bytes as f64 * self.nanos_per_byte) as u64;
+        // Clamp + advance in one atomic step: never let TAT fall behind `now`
+        // (no credit accumulation), then advance by the byte cost.
+        self.tat
+            .update(Ordering::Relaxed, Ordering::Relaxed, |old| {
+                old.max(now_nanos) + weight
+            });
```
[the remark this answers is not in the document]
jan-auer: That's fine, we compile with latest stable.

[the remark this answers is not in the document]
jan-auer: Zero is a nonsensical value so I left this gap in. Curious what other reviewers say, I'm happy to fix this to properly defined behavior.

objectstore-server/src/rate_limits.rs
```diff
@@ -252,97 +274,87 @@ impl RateLimiter {
...
+/// `burst_ns` nanoseconds ahead of the wall clock (burst tolerance). Recovery is
+/// continuous — no background tick required.
 #[derive(Debug)]
-struct EwmaEstimator {
-    accumulator: Arc<AtomicU64>,
-    estimate: Arc<AtomicU64>,
+struct BandwidthBucket {
+    /// Nanoseconds since `epoch` at which all recorded traffic is paid off.
+    ///
+    /// Saturates at `u64::MAX` after ~584 years of uptime.
```
lcian, line 324: Let's go.

objectstore-server/src/rate_limits.rs
lcian: Does that only apply to gauges, @jan-auer?
If not, I'm afraid that the `TimerGuard` I introduced might have this side effect too.

lcian (approved):
🚀

objectstore-server/src/rate_limits.rs
jan-auer: Just gauges, since those by design keep their old value until overwritten. Counters and distributions work differently.

objectstore-server/src/rate_limits.rs
```diff
@@ -252,97 +274,87 @@ impl RateLimiter {
...
+    burst_ns: u64,
 }
 
-impl EwmaEstimator {
-    fn new() -> Self {
+impl BandwidthBucket {
+    /// Creates a new bucket for the given rate and burst tolerance.
+    fn new(bps: u64, burst_ms: u64) -> Self {
+        let nanos_per_byte = 1_000_000_000.0 / bps as f64;
+        let burst_ns = burst_ms * 1_000_000;
```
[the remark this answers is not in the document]
jan-auer: That's acceptable.

PR #555, opened 2026-07-14
fix(server): Measure request duration end-to-end

lcian:
`server.requests.duration` now measures the request through response-body completion rather than stopping when a handler produces headers. This includes time spent streaming GET payloads, batch responses, and multipart-complete responses.

The middleware moves its timing guard into `MetricsBody`, which emits the metric only after Hyper has completed or dropped the response body. The wrapper preserves response metadata and existing `Content-Length` behavior.

Status tagging reflects how streaming ended:

- successful completion: the response status from the headers;
- client disconnect before completion: `499`;
- server-side stream error: `500`.

Completion handling is verified against Hyper 1.10.1 for empty/no-body responses, buffered bodies, ordinary streamed bodies, and terminal trailers. Regression tests cover each completion path, plus stream errors and interrupted streams.

Close FS-555

### DIFF
--- a/objectstore-server/docs/architecture.md
+++ b/objectstore-server/docs/architecture.md
@@ -76,20 +76,6 @@ A request flows through several layers before reaching the storage service:
    [`objectstore-types` docs](objectstore_types) for the header mapping) and
    the payload is streamed back.
 
-The `server.requests.duration` metric is measured **end-to-end**: the timing guard is
-moved into the response body, so it records the full request lifetime including the time
-spent streaming the response body back to the client (e.g. GET payloads, batch and
-multipart responses), not just the time to produce the response headers.
-
-The metric's `status` tag reflects the response status sent in the headers, **except** when
-the response body does not stream to completion:
-
-- **Client disconnect** mid-stream (the body is dropped before end-of-stream) is tagged
-  `499` (nginx' non-standard "client closed request").
-- **Server-side stream error** (the body yields an error while streaming) is tagged `500`.
-
-Both override the status sent in the headers.
-
 ## Authentication & Authorization
 
 Objectstore uses **JWT tokens with EdDSA signatures** (Ed25519) for
--- a/objectstore-server/docs/architecture.md
+++ b/objectstore-server/docs/architecture.md
@@ -76,6 +76,17 @@ A request flows through several layers before reaching the storage service:
    [`objectstore-types` docs](objectstore_types) for the header mapping) and
    the payload is streamed back.
 
+The `server.requests.duration` metric is measured **end-to-end**: the timing guard is
+moved into the response body, so it records the full request lifetime including the time
+spent streaming the response body back to the client (e.g. GET payloads, batch and
+multipart responses), not just the time to produce the response headers.
+
+The metric's `status` tag reflects the response status sent in the headers, **except** when
+the response body does not stream to completion, in which case it is tagged `499` (nginx'
+non-standard "client closed request"). This covers two intentionally conflated cases: the
+client disconnecting mid-stream, and a server-side error while streaming the body. Both are
+reported as `499`.
+
 ## Authentication & Authorization
 
 Objectstore uses **JWT tokens with EdDSA signatures** (Ed25519) for
--- a/objectstore-server/docs/architecture.md
+++ b/objectstore-server/docs/architecture.md
@@ -82,10 +82,13 @@ spent streaming the response body back to the client (e.g. GET payloads, batch a
 multipart responses), not just the time to produce the response headers.
 
 The metric's `status` tag reflects the response status sent in the headers, **except** when
-the response body does not stream to completion, in which case it is tagged `499` (nginx'
-non-standard "client closed request"). This covers two intentionally conflated cases: the
-client disconnecting mid-stream, and a server-side error while streaming the body. Both are
-reported as `499`.
+the response body does not stream to completion:
+
+- **Client disconnect** mid-stream (the body is dropped before end-of-stream) is tagged
+  `499` (nginx' non-standard "client closed request").
+- **Server-side stream error** (the body yields an error while streaming) is tagged `500`.
+
+Both override the status sent in the headers.
 
 ## Authentication & Authorization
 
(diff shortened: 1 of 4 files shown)


jan-auer:
> The `server.requests.duration` metric now records the full request lifetime

Changing `.duration` to cover the full time span makes sense. It can be useful to retain the old TTFB metric in addition to the full duration. We indirectly have those measurements through the server calls, but having a metric that covers all requests is more convenient.

objectstore-server/docs/architecture.md
```diff
@@ -76,6 +76,20 @@ A request flows through several layers before reaching the storage service:
    [`objectstore-types` docs](objectstore_types) for the header mapping) and
    the payload is streamed back.
 
+The `server.requests.duration` metric is measured **end-to-end**: the timing guard is
```
jan-auer: This kind of information is way too detailed for the high-level architecture docs. This can go either in the web/middleware module or onto the middleware that implements it.

objectstore-server/src/web/middleware.rs
```diff
@@ -111,26 +116,56 @@ pub async fn emit_request_metrics(mut request: Request, next: Next) -> Response
...
-    if let Some(guard) = guard {
-        guard.finish(response.status());
+    // Record the actual response status, then move the guard into the response body so the
+    // duration metric is emitted only when the body has finished streaming.
+    match guard {
+        Some(mut guard) => {
+            guard.finish(response.status());
+            let (parts, body) = response.into_parts();
+            let body = Body::new(MetricsBody::new(guard, body));
+            Response::from_parts(parts, body)
```
jan-auer: We can use `response.map()` here (and also for the sentry middleware)

objectstore-server/src/web/middleware.rs
```diff
@@ -111,26 +116,56 @@ pub async fn emit_request_metrics(mut request: Request, next: Next) -> Response
...
+        }
+        None => response,
     }
-    response
+}
+
+/// How the response body streaming ended, which determines the `status` tag on the emitted
+/// metric when it differs from the status sent in the headers.
+#[derive(Clone, Copy)]
+enum BodyOutcome {
```
jan-auer: Since these types are becoming verbose, I'd suggest to move them into a common module with `MetricsBody`.

objectstore-server/docs/architecture.md
lcian: Agreed. I didn't review the PR from Claude myself, otherwise would have probably removed this.

lcian:
bugbot run
@sentry review

objectstore-server/src/web/metrics_body.rs
```diff
@@ -0,0 +1,324 @@
+//! Response body wrapper that keeps a metrics guard alive until the body ends.
+//!
+//! Axum handlers produce response *headers* while the middleware stack unwinds,
+//! but the response *body* is streamed by hyper afterwards. To measure the true
+//! end-to-end request duration, the timing guard must live until the body has
+//! been fully streamed. [`MetricsBody`] owns an [`EmitMetricsGuard`] and drops
+//! it when the inner body reaches end-of-stream (or is dropped on client
+//! disconnect), at which point the guard emits `server.requests.duration`.
+//!
+//! When the body finishes, [`MetricsBody`] calls
```
lcian: This is kinda verbose and pretty detailed, but I think it's okay?

jan-auer: This is a too verbose, it essentially reads out the code. We should keep these docs to architecture, behavior, and guarantees. Additionally, this documents behavior of `MetricsBody`, so if we were to keep it, it would have to go on the struct docs instead. 

- Module-level docs are to describe entire concepts or parts of an application: the interplay between different components and the overall behavior
- Type-level docs are to describe its purpose and point at method-spanning usage
- Method-level docs then describe the actual behaivior of that method, incl. preconditions, errors, etc.

objectstore-server/src/web/metrics_body.rs
```diff
@@ -0,0 +1,324 @@
...
+use pin_project_lite::pin_project;
+use tokio::time::Instant;
+
+use crate::extractors::downstream_service::DownstreamService;
+
+/// How response-body streaming ended, which determines the `status` tag when it differs from
+/// the status sent in the headers.
+#[derive(Clone, Copy)]
+enum BodyOutcome {
+    /// The body was dropped before streaming completed, reported as `499`.
```
jan-auer: During the lifetime of `EmitMetricsGuard`, this acutally means that the body is still streaming. Only when the guard is dropped, we interpret this as a cancelled response.

objectstore-server/src/web/metrics_body.rs
```diff
@@ -0,0 +1,324 @@
...
+}
+
+impl Drop for EmitMetricsGuard {
+    fn drop(&mut self) {
+        let status = match self.outcome {
+            BodyOutcome::Completed => self.status.map(|status| status.as_u16()).unwrap_or(499),
+            BodyOutcome::Errored => 500,
+            BodyOutcome::Pending => 499,
+        }
+        .to_string();
```
jan-auer: _nit_: Move the `to_string()` into the metrics record below for nicer formatting.

objectstore-server/src/web/metrics_body.rs
```diff
@@ -0,0 +1,324 @@
...
+    /// Marks the response body stream as having errored server-side.
+    fn mark_errored(&mut self) {
+        self.outcome = BodyOutcome::Errored;
+    }
+}
+
+impl Drop for EmitMetricsGuard {
+    fn drop(&mut self) {
+        let status = match self.outcome {
+            BodyOutcome::Completed => self.status.map(|status| status.as_u16()).unwrap_or(499),
```
jan-auer: There is now a strange invariant between `outcome` and `status`. Technically, it's invalid to have both `BodyOutcome::Completed` and `status: None` at the same time, and it points at an implementation bug. 

A way to prevent that is:
- Create a `status` field on `MetricsBody`
- Change the variant to `BodyOutcome::Completed(StatusCode)`
- Bring back a single `finish()` method that we call in metrics body that moves the status from body to guard. No more separate `set_status` and `set_outcome`.

jan-auer (approved):

PR #556, opened 2026-07-15
feat: Read JWT from query params

lcian:
## Summary

Extends the Python client's `Session.object_url` to optionally produce a self-contained, token-carrying read URL, and teaches the server to accept the auth token from a query parameter so such URLs work.

- **Python client** (`clients/python`): `object_url(key, read_only_token_validity=timedelta | None)`. When `read_only_token_validity` is provided, it mints an `object.read`-only JWT scoped to the session's usecase and scope (via the existing `mint_token`), valid for that duration, base64url-encodes it, and appends it as the `os_auth` query parameter. Requires a `SecretKey`; raises `ValueError` otherwise. Default behavior (no arg) is unchanged.
- **Server** (`objectstore-server`): `from_token` now falls back to the `os_auth` query parameter when no `x-os-auth`/`Authorization` header token is present. The header takes precedence when both are present. Presigned URLs (`os_sig`) still take their own path.

## Design notes

- Query param name is `os_auth`, consistent with the existing `os_`-prefixed presign params (`os_sig`, `os_kid`, …).
- The JWT is already pre-base64-encoded, and `.`-separated.
- **Semantic difference from presigned URLs:** a read-only JWT authorizes reads on *any* object in the session's usecase+scope for its validity window (the `res` claim carries usecase+scope, not the object key), whereas a presigned URL binds the exact path+method. Documented in the docstring and README.
- `object_url`'s existing `http://host:port` URL building (scheme/IPv6 handling) is intentionally left unchanged — that's a separate fix.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/clients/python/src/objectstore_client/client.py
+++ b/clients/python/src/objectstore_client/client.py
@@ -442,9 +442,7 @@ class Session:
 
         return GetResponse(metadata, stream)
 
-    def object_url(
-        self, key: str, read_only_token_validity: timedelta | None = None
-    ) -> str:
+    def object_url(self, key: str, token_validity: timedelta | None = None) -> str:
         """
         Generates a GET url to the object with the given `key`.
 
@@ -452,27 +450,20 @@ class Session:
         NOTE however that the service does not strictly follow HTTP semantics,
         in particular in relation to `Accept-Encoding`.
 
-        When ``read_only_token_validity`` is provided, a read-only
-        (``object.read``) token scoped to this session's usecase and scope is
-        minted, valid for the given duration, and appended to the URL as the
-        base64url-encoded ``os_auth`` query parameter. The resulting URL is
-        self-contained: the recipient can fetch the object without supplying an
-        auth header.
+        When ``token_validity`` is provided, read-only authorization
+        information is embedded in the returned URL's query string, valid
+        for the given duration.
 
-        Note that unlike a pre-signed URL, which authorizes a single request on
-        one object, this token authorizes reads on any object in this session's
-        usecase and scope for the duration of its validity.
-
-        Raises ``ValueError`` if ``read_only_token_validity`` is provided but no
+        Raises ``ValueError`` if ``token_validity`` is provided but no
         ``SecretKey`` is configured on this session.
         """
         url = self._make_url(key, full=True)
-        if read_only_token_validity is None:
+        if token_validity is None:
             return url
 
         token = self.mint_token(
             permissions=[Permission.OBJECT_READ],
-            expiry_seconds=math.ceil(read_only_token_validity.total_seconds()),
+            expiry_seconds=math.ceil(token_validity.total_seconds()),
         )
         encoded = base64.urlsafe_b64encode(token.encode()).rstrip(b"=").decode()
         return f"{url}?{PARAM_AUTH}={encoded}"
--- a/clients/python/src/objectstore_client/client.py
+++ b/clients/python/src/objectstore_client/client.py
@@ -35,10 +35,6 @@ from objectstore_client.metrics import (
 from objectstore_client.multipart import MultipartUpload
 from objectstore_client.scope import Scope
 
-# Query parameter carrying a base64url-encoded JWT, mirroring the `x-os-auth`
-# header. Used by `Session.object_url` to embed an auth token in a URL.
-PARAM_AUTH = "os_auth"
-
 
 class GetResponse(NamedTuple):
     metadata: Metadata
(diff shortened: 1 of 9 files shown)


clients/python/src/objectstore_client/client.py
```diff
@@ -437,15 +442,40 @@ def get(
 
         return GetResponse(metadata, stream)
 
-    def object_url(self, key: str) -> str:
+    def object_url(
+        self, key: str, read_only_token_validity: timedelta | None = None
```
jan-auer: _nit_: This is a very long parameter name. How about `token_for` or something short that reads well?

clients/python/src/objectstore_client/client.py
```diff
@@ -437,15 +442,40 @@ def get(
...
+        url = self._make_url(key, full=True)
+        if read_only_token_validity is None:
+            return url
+
+        token = self.mint_token(
+            permissions=[Permission.OBJECT_READ],
+            expiry_seconds=math.ceil(read_only_token_validity.total_seconds()),
+        )
+        encoded = base64.urlsafe_b64encode(token.encode()).rstrip(b"=").decode()
+        return f"{url}?{PARAM_AUTH}={encoded}"
```
jan-auer: Calling out that this will break if `_make_url` were to include query params. Since it's an internal function and we have test coverage, fine to keep this way.

clients/python/tests/test_e2e.py
```diff
@@ -826,6 +826,44 @@ def test_presigned_get_encoding_corner_cases(server_url: str, key: str) -> None:
...
 
 
+def test_object_url_read_only_token_succeeds(server_url: str) -> None:
+    session = _presign_session(server_url)
+    session.put(b"read only hello", key="read-only-url")
+
+    url = session.object_url(
+        "read-only-url", read_only_token_validity=timedelta(minutes=5)
+    )
+    assert "os_auth=" in url
```
jan-auer: Could we run an actual URL parser here and take `os_auth` from parsed query params? That way we have coverage for potential future breakage, such as duplicate `?`, accidental wrong encoding, etc.

clients/python/tests/test_smoke.py
```diff
@@ -26,3 +37,30 @@ def test_object_url_with_base_path() -> None:
         session.object_url("foo/bar")
         == "http://127.0.0.1:8888/api/prefix/v1/objects/testing/org=12345;project=1337/foo/bar"
     )
+
+
+def test_object_url_read_only_token() -> None:
```
jan-auer: This and the above `object_url` tests aren't smoke tests and should better go into `test_e2e` or a unit test module.

objectstore-server/src/extractors/service.rs
```diff
@@ -107,6 +124,31 @@ fn has_signature(query: Option<&str>) -> bool {
...
+/// decode to a UTF-8 string.
+fn token_from_query(query: Option<&str>) -> Result<Option<String>, AuthError> {
+    let Some(value) = query.and_then(|query| {
+        query.split('&').find_map(|pair| {
+            let (key, value) = pair.split_once('=')?;
+            (key == AUTH_QUERY_PARAM).then_some(value)
+        })
+    }) else {
+        return Ok(None);
+    };
```
jan-auer: We're not URL-decoding here. Since we had issues with proxies re-encoding query strings, we should go the safe route and use a proper query string parser here to avoid future problems. 

Since this is being called from an extractor, there are two options.
1. Use the [`Query`](https://docs.rs/axum/latest/axum/extract/struct.Query.html) utility and extract it from parts in `AuthAwareService::from_token`. This requires a minimal `struct AuthParams { os_auth: Option<String> }`.
2. Use [`form_urlencoded::parse`](https://docs.rs/form_urlencoded/latest/form_urlencoded/fn.parse.html) directly and then iterate once to find our auth query param. (this is used internally by `Query`).

objectstore-server/tests/query_auth.rs
```diff
@@ -0,0 +1,144 @@
+//! End-to-end tests for the `os_auth` query parameter authentication path.
```
jan-auer, line 1: For a follow-up: These could all be unit tests. We tend to write many e2e tests, but few unit tests for the web layer. This makes tests slower and tests a rather large surface.

lcian: I think it may be a skill issue. I don't know how to write unit tests for axum endpoints, I need to look this up.

clients/python/src/objectstore_client/client.py
lcian: `token_validity`?

clients/python/tests/test_e2e.py
lcian: Done.

clients/python/tests/test_smoke.py
lcian: Removed as I don't think they were useful anyway, the e2e tests are enough.

jan-auer (approved):

PR #559, opened 2026-07-15
ref(service): Remove details from GCS error responses

jan-auer:
GCS optionally includes a `<Details>` element that contains additional info. In
practice, this turns out to be internal references that we cannot use, but that
interfere with our issue grouping. 

This removes the element entirely.

Ref FS-439


lcian (approved):
Looks good to me.

PR #560, opened 2026-07-16
feat(stresstest): Optionally emit metrics

jan-auer:
Records metrics for throughput, bandwidth, and object size. This allows us to
plot the same metrics that stresstest outputs on stdout in our dashboards for
historical records.

Configuration is the same as for objectstore server.

Closes FS-77

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4549,6 +4549,7 @@ dependencies = [
  "humantime-serde",
  "indicatif",
  "objectstore-client",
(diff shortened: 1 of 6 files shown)


lcian (approved):
Looks good to me as initial addition. Maybe you could double double check batch mode and multipart mode next, because I think we might misreport some metrics in those cases.

jan-auer:
Good point, will follow up with more stresstest metrics in the future

PR #561, opened 2026-07-16
feat: Add bulk-upload util

lcian:
This is a utility that mirrors the `sentry-cli snapshots upload` command for easy sandbox tests.
It can be used in sandbox with the `preprod` manifest committed here: https://github.com/getsentry/terraform-sandboxes.private/pull/513/changes.
Tune the parameters, then run `kubectl apply -f tests/preprod/manifest.yaml`.
The snapshots are already mounted at `/data/preprod`, so no need to do anything there. They've been generated with a distribution having p50 == 10 KiB, p99 == 215 KiB.

You likely want to tune the parameters `limit` (to not work on all the 40k files, just a subset), `head_first`, `individual_concurrency`, `batch_concurrency` (if you don't set the last 2, the defaults from `objectstore_client` will be used, which matches what Sentry CLI uses).

The `prefix` parameter can also be useful when performing multiple tests: the first run will populate the objects in Objectstore; if you don't set a `prefix`, you will (head and) override existing objects, which might result in a slightly different benchmark from fresh inserts.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -535,6 +535,19 @@ dependencies = [
 ]
 
 [[package]]
+name = "bulk-upload"
+version = "0.1.0"
+dependencies = [
+ "anyhow",
+ "argh",
+ "futures",
+ "indicatif",
+ "objectstore-client",
+ "tikv-jemallocator",
+ "tokio",
+]
+
+[[package]]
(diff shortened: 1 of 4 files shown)


lcian:
Not needed.

PR #547, opened 2026-07-09
ref(service): Replace Error with anyhow-like struct with embedded kind

lcian:
Replaces the `objectstore_service::error::Error` enum with a struct containing an independent `ErrorKind` classification and an `anyhow::Error` diagnostic.

This separates error classification from error origin: anyhow preserves the source and stacked context chain, while backends can reclassify errors with `.kind()` and add context with `.context()` without `map_err` boilerplate. The API maps the flat `ErrorKind` to HTTP status codes.

Closes FS-358
Ref FS-439

PR #562, opened 2026-07-16
ref(server): Introduce a queue to the concurrency limiter

jan-auer:
The concurrency limiter can now absorb brief capacity bursts by queuing requests instead of rejecting them immediately. This is implemented through semaphores: an outer semaphore (capacity `max + queue`) gates total participants, and an inner semaphore (capacity `max`) gates execution. Requests that pass the outer gate wait up to a configurable timeout for an execution permit; requests that exceed both limits are rejected with `AtCapacity` as before.

With the default `concurrency_queue = 0`, both semaphores have the same capacity and behavior is identical to the current immediate-reject model. In this case, there is no queuing and no added latency.

Two new config fields control the queue. Sizing guidance: set the queue depth to roughly `permit_release_rate × acceptable_added_latency`, and keep the timeout comfortably below client request timeouts.

```yaml
service:
  max_concurrency: 500
  concurrency_queue: 100
  concurrency_queue_timeout: 1s
```

Batch `stream()` reservations continue to use non-blocking `try_acquire_many` and do not enter the queue.

Ref FS-447

### DIFF
--- a/objectstore-server/docs/architecture.md
+++ b/objectstore-server/docs/architecture.md
@@ -232,9 +232,9 @@ Direct 503 rejection is preferred over readiness-based backpressure:
 Beyond rate limiting and the web concurrency limit, the
 [`StorageService`](objectstore_service::StorageService) enforces a second
 layer of backpressure through a concurrency limit on in-flight backend
-operations, configured via `service.max_concurrency`. When exceeded, requests
-receive HTTP 429. See the [service architecture docs](objectstore_service) for
-details.
+operations, configured via `service.max_concurrency` and related options.
+When exceeded, requests receive HTTP 429. See the [service architecture
+docs](objectstore_service) for details.
 
 ## KEDA Metrics
 
--- a/objectstore-server/src/config.rs
+++ b/objectstore-server/src/config.rs
@@ -508,6 +508,8 @@ pub struct Config {
 /// # Environment Variables
 ///
 /// - `OS__SERVICE__MAX_CONCURRENCY`
+/// - `OS__SERVICE__CONCURRENCY_QUEUE`
+/// - `OS__SERVICE__CONCURRENCY_QUEUE_TIMEOUT`
 #[derive(Debug, Deserialize, Serialize)]
 #[serde(default)]
 pub struct Service {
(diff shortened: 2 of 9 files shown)


objectstore-service/src/concurrency.rs
```diff
@@ -23,65 +23,126 @@ const EMITTER_INTERVAL: Duration = Duration::from_secs(1);
...
+            .await
+            .map_err(|_| Error::AtCapacity)?
+            .map_err(|_| Error::AtCapacity)?;
+
+        Ok(ConcurrencyPermit {
+            task_permit: Some(task_permit),
+            queue_permit: Some(queue_permit),
+            released: Arc::clone(&self.released),
+        })
+    }
```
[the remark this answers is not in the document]
jan-auer: Concurrency 0 is a nonsensical value, but this is easy to guard against.

objectstore-service/src/concurrency.rs
```diff
@@ -21,67 +21,145 @@ use crate::error::{Error, Result};
...
+        let task_permit = self
+            .tasks
             .clone()
             .try_acquire_many_owned(count)
             .map_err(|_| Error::AtCapacity)?;
+
         Ok(ConcurrencyPermit {
-            permit: Some(permit),
+            task_permit: Some(task_permit),
+            queue_permit: Some(queue_permit),
```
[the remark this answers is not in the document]
jan-auer: That's an inherent limit of the design. We will solve this in a follow-up by removing `acquire_many` and changing batch-processing so it polls individually.

objectstore-service/src/concurrency.rs
```diff
@@ -21,67 +21,145 @@ use crate::error::{Error, Result};
...
+            .queue
+            .clone()
+            .try_acquire_owned()
+            .map_err(|_| Error::AtCapacity)?;
+
+        let acquire = self.tasks.clone().acquire_owned();
+        let task_permit = tokio::time::timeout(self.timeout, acquire)
+            .await
+            .map_err(|_| Error::AtCapacity)?
+            .map_err(|_| Error::AtCapacity)?;
```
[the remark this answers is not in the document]
jan-auer: Default in config is 1s, using 0s makes no sense.

objectstore-server/docs/architecture.md
```diff
@@ -232,9 +232,9 @@ Direct 503 rejection is preferred over readiness-based backpressure:
 Beyond rate limiting and the web concurrency limit, the
 [`StorageService`](objectstore_service::StorageService) enforces a second
 layer of backpressure through a concurrency limit on in-flight backend
-operations, configured via `service.max_concurrency`. When exceeded, requests
-receive HTTP 429. See the [service architecture docs](objectstore_service) for
-details.
+operations, configured via `service.max_concurrency` and related options.
+When exceeded, requests receive HTTP 429. See the [service architecture
```
lcian, line 236: Should this be a 503 instead? It's technically not a rate limit, we just happen to be unavailable at that precise time.

lcian: In https://github.com/getsentry/objectstore/pull/547, I change the mapping of this to 503.

lcian (approved):
Looks good to me.

objectstore-server/src/config.rs
```diff
@@ -521,12 +523,39 @@ pub struct Service {
...
+    /// `concurrency_queue_timeout`) instead of being rejected immediately.
+    /// Requests beyond that are rejected with HTTP 429.
+    ///
+    /// Sizing guidance: `concurrency_queue ≈ permit_release_rate ×
+    /// acceptable_added_latency`.
+    ///
+    /// # Default
+    ///
+    /// `0`
+    pub concurrency_queue: u32,
```
lcian, line 540:
```suggestion
    pub concurrency_queue_size: u32,
```
Maybe?

jan-auer: I'm also not entirely happy with this. Was also thinking of `max_queued`. Let's wait for the follow-up that introduces another parameter for batch requests and then reassess.

PR #565, opened 2026-07-21
ref(bigtable): move payload column to new "fpg"/"fpm" column families

matt-codecov:
RFC

the COGS rollup query we have to break down our Bigtable usage is very slow in large regions (exceeds BigQuery's 6h cutoff) and has significant impact on production latencies while it's running. this PR is a structural change to our Bigtable schema that would solve those problems.

the query groups by usecase and will sum the sizes of:
- the `m` column and the payload (via `m.size`, https://github.com/getsentry/objectstore/pull/490) for objects
- the `t` and `r` columns for tombstones

the query is slow because, as i understand it, the `p` column is in the same column family which puts it in the same SSTable. so even though the query doesn't actually read the `p` column, we still have to seek past it, and our query consequently tries to monopolize the disk head for hours for a front-to-back scan.

by moving `p` to separate `fpg`/`fpm` families (payload versions of `fg`/`fm`), queries that don't touch `p` only have to seek over `m`/`t`/`r` which is like 1% of the data

incidentally this may improve the performance of production operations that don't touch the payload (`get_metadata()`)

### DIFF
--- a/devservices/run-bigtable.sh
+++ b/devservices/run-bigtable.sh
@@ -14,7 +14,8 @@ for _ in {1..4}; do
 done
 
 cbt deletetable objectstore || true
-cbt createtable objectstore families=fg,fm
+cbt createtable objectstore families=fg,fm,fpg,fpm
 cbt setgcpolicy objectstore fg maxage=1s
+cbt setgcpolicy objectstore fpg maxage=1s
 
 wait $EMULATOR_PID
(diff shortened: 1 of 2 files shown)


jan-auer:
The metadata and payload columns are in the same column family to ensure they are deleted together on expiry. In the [schema design guidelines](https://docs.cloud.google.com/bigtable/docs/schema-design), Google specifically mentions to put related columns into the same family. 

If we separate them now, we have to watch very closely for unintended consequences: Performance, atomicity, consistency. This is not a simple change.

I would advise against doing this.

jan-auer:
Closing as we're going to use a different approach now.

PR #567, opened 2026-07-22
fix(python): Use pool scheme and bracket IPv6 hosts in object URLs

jan-auer:
`Session.object_url()` hardcoded the `http://` scheme and did not bracket IPv6 hosts, producing wrong or malformed URLs for HTTPS base URLs and IPv6 addresses. The sibling `presigned_object_url()` handled both correctly.

A shared `_base_url()` helper now reconstructs the origin from the pool's scheme, host, and port with proper IPv6 bracketing. Both `_make_url` and `presigned_object_url` use it, eliminating the duplication that caused the bug.

Fixes FS-430

### DIFF
--- a/clients/python/src/objectstore_client/client.py
+++ b/clients/python/src/objectstore_client/client.py
@@ -272,11 +272,19 @@ class Session:
             headers["x-os-auth"] = f"Bearer {token}"
         return headers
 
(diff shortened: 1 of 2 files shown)


lcian (approved):
Looks good to me.

PR #568, opened 2026-07-22
fix(python): Align empty-scope URL rendering with Rust client

jan-auer:
The Python client rendered empty scopes as an empty string, producing `/v1/objects/{usecase}//{key}`, while the Rust client and server use the `_` sentinel, producing `/v1/objects/{usecase}/_/{key}`. An object written with one client under an empty scope was not addressable by the other.

This also fixes a bug where falsy scope values like `0` or `False` were rejected as "empty" because the truthiness check ran before `str()` conversion.

Fixes FS-435


lcian (approved):
Looks good to me.

PR #569, opened 2026-07-22
fix(service): Filter expired objects in the S3-compatible backend

jan-auer:
The GCS and BigTable backends filter out expired objects at read time — returning `None` when the stored expiry timestamp is in the past. The S3-compatible backend was missing this check, so TTL and TTI objects remained readable past their deadline until the storage lifecycle rule physically deleted them (up to ~48h later).

This lifts the `expire_at` parsing out of the TTI-only block and adds the same expiry guard before the TTI bump, matching the GCS pattern. The response body is drained before returning `None` to avoid connection leaks.

Ref FS-403

### DIFF
--- a/objectstore-service/src/backend/s3_compatible.rs
+++ b/objectstore-service/src/backend/s3_compatible.rs
@@ -438,6 +438,7 @@ mod tests {
     }
 
     #[tokio::test]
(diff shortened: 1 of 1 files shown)


lcian (approved):
Looks good to me.

PR #570, opened 2026-07-22
fix(python): Return None from get() on 404 to match Rust client

jan-auer:
`Session.get()` now returns `None` when the object does not exist, matching
the behavior of `head()` and the Rust client. Previously it raised
`RequestError` with status 404, forcing callers to catch and inspect the
exception for the common not-found case.

This is a breaking change for callers that catch `RequestError` to detect
missing objects — they should check for `None` instead.

Fixes FS-427


lcian (approved):
Looks good to me, this is pretty good as it also forces the caller to handle the 404 instead of getting a possibly unhandled exception.

PR #575, opened 2026-07-24
fix(gcs): Retry multipart operations

lcian:
Retryable GCS multipart operations now go through the same request-level retry helper already used by the JSON API path, so transient transport/`5xx` failures are absorbed inside the backend instead of bouncing to callers.

Supersedes https://github.com/getsentry/objectstore/pull/566
Closes FS-450
Fixes OBJECTSTORE-3XB

### DIFF
--- a/objectstore-service/src/backend/gcs.rs
+++ b/objectstore-service/src/backend/gcs.rs
@@ -886,6 +886,11 @@ impl From<XmlError> for crate::multipart::CompleteMultipartError {
 /// XXX: Any change that affects this implementation should be manually tested against real GCS.
 /// That's because the fork of [storage-testbench](https://github.com/googleapis/storage-testbench)
 /// that we test against has an incomplete implementation of the XML multipart API that likely doesn't match GCS's behavior in many cases.
+///
+/// Request-level retries use the same internal retry helper as JSON API calls for
+/// initiate/list/abort. Part upload is intentionally not retried: the body is streamed and not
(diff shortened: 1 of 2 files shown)


objectstore-service/src/backend/gcs.rs
```diff
@@ -1026,16 +1044,32 @@ impl MultipartUploadBackend for GcsBackend {
...
+                let resp = self
+                    .request(Method::DELETE, url)
+                    .await?
+                    .send_traced()
+                    .await
+                    .map_err(|e| Error::reqwest("GCS: abort multipart upload", e))?;
 
-        Ok(())
+                // Idempotent: absent upload means the postcondition is already satisfied
+                // (including retry after a successful abort whose response was lost).
```
jan-auer: Could you double-check this? When testing `complete` recently, I found that GCS responds with 200 on repeat calls rather than with 404. Maybe it's the same here.

Of course, there will be a limit to this. After some time, the upload will be cleaned up.

jan-auer (approved):
LGTM. Please add `complete`, too.

objectstore-service/src/backend/gcs.rs
lcian: Checking.

lcian: Indeed it's 200 on complete and 204 on abort. Changing this.

lcian: Okay just checked and fixed. Also retrying complete now.

PR #581, opened 2026-07-31
feat(types): Support Unicode in object metadata headers

jan-auer:
Object metadata can now carry arbitrary Unicode. Previously a non-ASCII value failed the whole request with `400 Bad Request` and the message "error dealing with http headers".  Headers only accept visible ASCII, so unicode was unstorable by any route.

Metadata values are now percent-encoded at the transport layer. `Metadata` itself is unchanged: it holds logical `String`s throughout, and escaping exists only where a value is written into a header. Additionally, non-ASCII filenames additionally get an RFC 8187 `filename*` parameter on `Content-Disposition`, with the existing quoted-string kept as the ASCII fallback.

### Wire compatibility

Valid ASCII bytes pass through untouched, so values that are already plain ASCII are byte-identical to what they were before and existing clients see no change. The one visible difference is a value containing a literal `%`, which is now escaped.

The batch operation key header, which had its own hand-rolled `NON_ALPHANUMERIC` encoding in three places across two crates, now shares this set.

### Backends

Both header-based backends (GCS and S3) now store the escaped form.

GCS escapes on **both** of its write paths, including the JSON one that could carry Unicode verbatim. It follows Google's own advice: "you should generally avoid non-ascii characters, because they are not permitted in HTTP headers, which the XML API uses." The UTF-8-preserving behavior is undocumented, and raw UTF-8 additionally loses leading whitespace to header trimming and silently rewrites invalid UTF-8.

### Not covered

Custom metadata **keys** are still ASCII-only and still lowercased by the header map, so `MyKey` round-trips as `mykey` as it always has. Escaping keys would work but changes the wire form of every existing key without fixing the case-folding, so it is left alone.

Fixes FS-467


lcian (approved):
Looks good to me.

PR #582, opened 2026-08-01
feat: keeper trait & sqlite-backed implementation

aldy505:
[Internal Slack thread](https://sentry.slack.com/archives/C0BLLQP4H6E/p1785331379027609).

This is required for filesystem & S3-compatible API backends. Later, I'll try to integrate this with filesystem backend, and create Postgres-backed keeper. 

Refs FS-482

### DIFF
--- a/migrations/sqlite/0001_keeper.sql
+++ b/migrations/sqlite/0001_keeper.sql
@@ -2,6 +2,6 @@ CREATE TABLE IF NOT EXISTS ttl_keeper (
     object_id TEXT PRIMARY KEY,
     expiration_policy INTEGER NOT NULL DEFAULT 0,
     duration INTEGER,
-    created_at INTEGER NOT NULL,
-    expires_at INTEGER
+    time_created INTEGER NOT NULL,
+    time_expires INTEGER
 );
--- a/migrations/sqlite/0001_keeper.sql
+++ b/migrations/sqlite/0001_keeper.sql
@@ -1,0 +1,7 @@
+CREATE TABLE IF NOT EXISTS ttl_keeper (
+    object_id TEXT PRIMARY KEY,
+    expiration_policy INTEGER NOT NULL DEFAULT 0,
+    duration INTEGER,
+    created_at INTEGER NOT NULL,
+    expires_at INTEGER
+);
--- a/objectstore-server/Cargo.toml
+++ b/objectstore-server/Cargo.toml
@@ -37,17 +37,40 @@ objectstore-types = { workspace = true }
 papaya = { workspace = true }
 pin-project-lite = { workspace = true }
 rand = { workspace = true }
-reqwest = { workspace = true, features = ["charset", "http2", "system-proxy", "native-tls-no-alpn"] }
+reqwest = { workspace = true, features = [
+    "charset",
+    "http2",
+    "system-proxy",
+    "native-tls-no-alpn",
+] }
 rustls = { workspace = true }
 secrecy = { workspace = true, features = ["serde"] }
-sentry = { workspace = true, features = ["tower-axum-matched-path", "tracing", "logs"] }
+sentry = { workspace = true, features = [
+    "tower-axum-matched-path",
+    "tracing",
+    "logs",
+] }
 serde = { workspace = true }
 serde_json = { workspace = true }
 thiserror = { workspace = true }
 thread_local = { workspace = true }
-tokio = { workspace = true, features = ["fs", "macros", "net", "rt-multi-thread", "signal", "sync", "time"] }
+tokio = { workspace = true, features = [
+    "fs",
+    "macros",
+    "net",
+    "rt-multi-thread",
+    "signal",
+    "sync",
+    "time",
+] }
 tower = { workspace = true }
-tower-http = { workspace = true, features = ["catch-panic", "metrics", "set-header", "trace"] }
+tower-http = { workspace = true, features = [
+    "catch-panic",
+    "metrics",
+    "set-header",
+    "trace",
+] }
+sqlx.workspace = true
 
 [target.'cfg(target_os = "linux")'.dependencies]
 jemalloc_pprof = { workspace = true, optional = true }
--- a/objectstore-server/Cargo.toml
+++ b/objectstore-server/Cargo.toml
@@ -37,40 +37,17 @@ objectstore-types = { workspace = true }
 papaya = { workspace = true }
 pin-project-lite = { workspace = true }
 rand = { workspace = true }
-reqwest = { workspace = true, features = [
-    "charset",
-    "http2",
-    "system-proxy",
-    "native-tls-no-alpn",
-] }
+reqwest = { workspace = true, features = ["charset", "http2", "system-proxy", "native-tls-no-alpn"] }
 rustls = { workspace = true }
 secrecy = { workspace = true, features = ["serde"] }
-sentry = { workspace = true, features = [
-    "tower-axum-matched-path",
-    "tracing",
-    "logs",
-] }
+sentry = { workspace = true, features = ["tower-axum-matched-path", "tracing", "logs"] }
 serde = { workspace = true }
 serde_json = { workspace = true }
 thiserror = { workspace = true }
 thread_local = { workspace = true }
-tokio = { workspace = true, features = [
-    "fs",
-    "macros",
-    "net",
-    "rt-multi-thread",
-    "signal",
-    "sync",
-    "time",
-] }
+tokio = { workspace = true, features = ["fs", "macros", "net", "rt-multi-thread", "signal", "sync", "time"] }
 tower = { workspace = true }
-tower-http = { workspace = true, features = [
-    "catch-panic",
-    "metrics",
-    "set-header",
-    "trace",
-] }
-sqlx.workspace = true
+tower-http = { workspace = true, features = ["catch-panic", "metrics", "set-header", "trace"] }
 
 [target.'cfg(target_os = "linux")'.dependencies]
 jemalloc_pprof = { workspace = true, optional = true }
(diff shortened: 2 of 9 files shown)


objectstore-server/Cargo.toml
```diff
@@ -38,17 +38,40 @@ papaya = { workspace = true }
 pin-project-lite = { workspace = true }
 percent-encoding = { workspace = true }
 rand = { workspace = true }
-reqwest = { workspace = true, features = ["charset", "http2", "system-proxy", "native-tls-no-alpn"] }
+reqwest = { workspace = true, features = [
```
jan-auer: Let's undo these formatting changes to keep dependencies in a single line.  Same for other Cargo.toml files.

objectstore-server/Cargo.toml
```diff
@@ -38,17 +38,40 @@ papaya = { workspace = true }
...
+] }
 tower = { workspace = true }
-tower-http = { workspace = true, features = ["catch-panic", "metrics", "set-header", "trace"] }
+tower-http = { workspace = true, features = [
+    "catch-panic",
+    "metrics",
+    "set-header",
+    "trace",
+] }
+sqlx.workspace = true
```
jan-auer: Looks like this is unused in the server.

objectstore-service/Cargo.toml
```diff
@@ -25,10 +25,20 @@ objectstore-metrics = { workspace = true }
...
+    "json",
+    "multipart",
+    "native-tls-no-alpn",
+    "stream",
+    "system-proxy",
+] }
 sentry = { workspace = true }
 serde = { workspace = true }
 serde_json = { workspace = true }
+sqlx = { workspace = true }
```
jan-auer: Ideally, leave the workspace dependency without most features and include the required features here. This way, when we use conditional compilation of certain sub-crates we automatically get the smallest possible set of features.

objectstore-service/src/keeper/mod.rs
```diff
@@ -0,0 +1,34 @@
...
+/// Represents the computed expiry information for an object.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct ObjectExpiry<'a> {
+    id: &'a ObjectId,
+    expiration_policy: ExpirationPolicy,
+}
+
+/// Object retention keeper trait.
+#[async_trait::async_trait]
+pub trait Keeper: Send + Sync {
```
jan-auer, line 21: Can we find a more descriptive name for this? `Keeper` is a nice and short name, but it lacks context on what this is for. 

Intuitively, what we're building is GC, so I'm throwing that in as a suggestion.

objectstore-service/src/keeper/mod.rs
```diff
@@ -0,0 +1,34 @@
...
+    /// Keep is the first step in the object retention lifecycle.
+    /// Practically speaking, it would not be kept if the `expiration_policy` is `Manual`.
+    /// The `expiration_policy` is set by the client at upload time via the supplied Metadata.
+    async fn keep(&self, id: &ObjectId, expiration_policy: ExpirationPolicy) -> Result<()>;
+
+    /// Remove is the final step in the object retention lifecycle.
+    /// It is called by a cleanup worker when the object is no longer needed.
```
jan-auer, line 28: Either this, or when the object is deleted per user request (or as part of tiered storage cleanup, but that's the same method).

objectstore-service/src/keeper/mod.rs
```diff
@@ -0,0 +1,34 @@
...
+    async fn remove(&self, id: &ObjectId) -> Result<()>;
+
+    /// Marks an object as accessed. For `expiration_policy` of `TimeToIdle`, this will
+    /// extend the object retention.
+    async fn mark_accessed(&self, id: &ObjectId) -> Result<()>;
```
jan-auer: Can we instead make this more of an "update" method and pass an explicit new expiration time?

I'm currently working on refactoring TTI to centralize its logic. Right now, backends have to handle this all internally, which leads to multiple problems. The biggest one is that the eviction timestamp can go out of sync with the actual object. 

We can fix that by passing an explicit expiration time around.

objectstore-service/src/keeper/sqlite_backed.rs
```diff
@@ -0,0 +1,456 @@
...
+    /// The storage path key for the object.
+    pub object_id: String,
+    /// Encoded expiration policy: 0 = Manual, 1 = TimeToLive, 2 = TimeToIdle.
+    pub expiration_policy: i32,
+    /// The expiration duration in seconds, if applicable.
+    pub duration: Option<i64>,
+    /// Unix timestamp (seconds) when the row was created.
+    pub created_at: i64,
+    /// Unix timestamp (seconds) when the object expires, if applicable.
+    pub expires_at: Option<i64>,
```
jan-auer: _nit_: Where possible, let's adopt the same terminology we also use in `Metadata`, such as `time_expires`, etc.

jan-auer (commented):
Thank you! This is a first review pass with some questions.

The design looks good, specifically:
 - It's great to introduce a trait so we can have different implementations of this via config. 
 - The overall keeper interface is simple and doesn't assume too much about object lifecycle (though see my comment on TTI below).
 - It makes sense that the backend "owns" the instance of the keeper.

Some question to the overall design:
 1. Who is responsible to scan for deleted objects and drive deletion and what will the interface for this look like?
    - If it is the keeper, how does it tell the backend to delete?
    - If it is the backend, how does it use the keeper interface to scan? There's no method to iterate objects that are ready for deletion
 2. If the keeper database gets corrupted, deleted, or the keeper is swapped out, we lose all information on objects and GC for those objects will no longer happen. Doesn't have to be solved immediately, but do you already have thoughts on this? 
 3. How will the keeper ensure that on concurrent writes to the same key that keeper's entry corresponds to what is stored in the backend?
    - Two requests PUT at the same time with different expires_at. Only one of them will win, keeper must end up with the same expiry time.
    - One request PUTs and one DELETEs at the same time. One of them wins, and the keeper must match. 

When calling `keep`, ensure to do so _before_ writing the object. Otherwise, we could end up with a persisted object but without keeper entry.

sqlite can only have a single writer attached to a database file at any time. This means when sqlite is used, one cannot run multiple objectstore instances. This is an important restriction we should add to some doc comment and later to the config that exposes this.

migrations/sqlite/0001_keeper.sql
```diff
@@ -0,0 +1,7 @@
+CREATE TABLE IF NOT EXISTS ttl_keeper (
+    object_id TEXT PRIMARY KEY,
+    expiration_policy INTEGER NOT NULL DEFAULT 0,
+    duration INTEGER,
+    created_at INTEGER NOT NULL,
+    expires_at INTEGER
```
jan-auer: We will likely need an index on this.

aldy505:
>  - Who is responsible to scan for deleted objects and drive deletion and what will the interface for this look like?
>     - If it is the keeper, how does it tell the backend to delete?
>     - If it is the backend, how does it use the keeper interface to scan? There's no method to iterate objects that are ready for deletion

I'm thinking of a separate cleanup process for this. Therefore only the keeper is the one who's responsible.

>  - If the keeper database gets corrupted, deleted, or the keeper is swapped out, we lose all information on objects and GC for those objects will no longer happen. Doesn't have to be solved immediately, but do you already have thoughts on this? 

I haven't think this through. On my current proposal, I know that there will be lots of orphan objects, and there's no way to figure out whether it should be managed by the keeper or not. 

One thing that came across my mind just now is to append it on the sidecar metadata file (that we've talked about on Slack). I'm thinking it only for a last resort recovery option. 


> sqlite can only have a single writer attached to a database file at any time. This means when sqlite is used, one cannot run multiple objectstore instances. This is an important restriction we should add to some doc comment and later to the config that exposes this. 

Oh yes, and I'm considering to add Postgres as another keeper backend.

objectstore-server/Cargo.toml
aldy505: This is me toying around with the sqlx migrate stuff, to have compile time checks for the query.

objectstore-service/src/keeper/mod.rs
aldy505: I took Zookeeper and ClickHouse Keeper as a reference for this 😄

aldy505:
>  3. How will the keeper ensure that on concurrent writes to the same key that keeper's entry corresponds to what is stored in the backend?
> 
>     - Two requests PUT at the same time with different expires_at. Only one of them will win, keeper must end up with the same expiry time.
> 
>     - One request PUTs and one DELETEs at the same time. One of them wins, and the keeper must match. 

Yes I'm aware of this. Since sqlite is a single writer, I would trust whoever enters objectstore first.

jan-auer:
> Since sqlite is a single writer, I would trust whoever enters objectstore first.

Sqlite is a single writer, but objectstore allows non-blocking concurrent requests on the same object. Since these requests consist of several sequential operations, there can be races like TOCTOU and lost updates. A request that comes in first may not be the first to finish, and there can be any form of interleaving. 

In principle, there are these options:
 - Synchronize and block. This is what we do in the in-memory backend for testing. We do not do this in other backends because we lack the primitives for synchronization across multiple instances of objectstore. Also, this would further increase latency.
 - Check and fail if there is an operation ongoing. Again, we're lacking the primitives here in many cases. 
 - Update optimistically. Treat concurrent operations like they were serialized and ensure there is consistent outcome. This requires a form of atomic updates or CAS.

objectstore-service/src/keeper/mod.rs
aldy505: Therefore we need to trust whatever the server is giving to the keeper?

aldy505:
@jan-auer @lcian Generic question: what to do when the expiration duration is a negative number? Assuming the expiration duration is `i64` (not `u64`)

jan-auer:
> @jan-auer @lcian Generic question: what to do when the expiration duration is a negative number? Assuming the expiration duration is `i64` (not `u64`)

In metadata, the duration for expiring policies is always positive (see [`ExpirationPolicy`](https://getsentry.github.io/objectstore/rust/objectstore_client/enum.ExpirationPolicy.html), `Duration` cannot be negative). On top of that, the service is always computing an absolute `time_expires` which is a unix timestamp. 

Is there a place where you're seeing negative durations (other than for expired objects, of course)?

PR #584, opened 2026-08-03
fix(gcs): Swallow TTI bump race errors and use preconditions

lcian:
With this PR we use the `generation` and `metageneration` returned by GCS as update preconditions when bumping TTI.
We should probably do this in all methods to guarantee consistency when we're using GCS as a standalone backend, but for now we just do it on TTI bumps.

This turns the `409 Conflict` errors we were getting when two metadata updates were racing in `412 Precondition Failed` errors, which we just swallow and treat as success when bumping TTI.

### DIFF
--- a/objectstore-service/src/backend/gcs.rs
+++ b/objectstore-service/src/backend/gcs.rs
@@ -110,6 +110,10 @@ struct GcsObject {
     )]
     pub custom_time: Option<SystemTime>,
 
+    /// Version of the object's contents.
+    #[serde(skip_serializing)]
+    pub generation: String,
+
     /// The `Content-Length` of the data in bytes. GCS returns this as a string.
     ///
     /// GCS sets this in metadata responses. We can use it to know the size of an object
@@ -141,6 +145,7 @@ impl GcsObject {
             size: metadata.size.map(|size| size.to_string()),
             content_encoding: None,
             custom_time: None,
+            generation: String::new(),
             time_created: metadata.time_created,
             metageneration: String::new(),
             metadata: BTreeMap::new(),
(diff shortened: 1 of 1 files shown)


objectstore-service/src/backend/gcs.rs
```diff
@@ -571,33 +584,53 @@ impl GcsBackend {
...
+        metageneration: &str,
+    ) -> Result<()> {
         #[derive(Debug, Serialize)]
         #[serde(rename_all = "camelCase")]
         struct CustomTimeRequest {
             #[serde(with = "humantime_serde")]
             custom_time: SystemTime,
         }
 
+        let object_url = with_metageneration_match(object_url, metageneration);
```
jan-auer: We call `with_metageneration_match ` only here, I'd suggest to inline for readability.

objectstore-service/src/backend/gcs.rs
```diff
@@ -571,33 +584,53 @@ impl GcsBackend {
...
+                    .await?
+                    .drain_body()
+                    .await;
+                Ok(())
+            }
+            .await;
+
+            // Bumping TTI is opportunistic. A concurrent metadata writer won the CAS race,
+            // so leave its update intact and let a future read evaluate the TTI again.
+            match result {
```
jan-auer: You can match this directly on the result of `check_error`, so then we can avoid wrapping the above request call into a nested `async` block.

jan-auer (approved):

objectstore-service/src/backend/gcs.rs
lcian: Done — the single-use URL helper is now inlined at the update site.

*— Claude Code*

objectstore-service/src/backend/gcs.rs
lcian: Done — the update now matches directly on the result of , without a nested async block.

*— Claude Code*

lcian: Correction: the update now matches directly on the checked response, without a nested async block.

*— Claude Code*

objectstore-service/src/backend/gcs.rs
```diff
@@ -127,6 +127,14 @@ struct GcsObject {
...
     #[serde(default, skip_serializing_if = "BTreeMap::is_empty")]
     pub metadata: BTreeMap<GcsMetaKey, String>,
+
+    /// Version of the object's contents.
+    #[serde(skip_serializing)]
+    pub generation: String,
+
+    /// Version of the object's metadata.
+    #[serde(skip_serializing)]
+    pub metageneration: String,
```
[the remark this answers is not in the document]
lcian: They're always present.

PR #586, opened 2026-08-04
feat(clients): Allow marking uploads as precompressed

jan-auer:
Uploading already-compressed data and having it recorded as compressed was impossible: `compression` on a put both set the `Content-Encoding` metadata and told the client to compress, so the only way to skip compression also dropped the encoding. Objects then came back as raw compressed bytes that no client could transparently decompress.

Both clients gain a way to declare that the payload is already compressed. The bytes are uploaded verbatim and the algorithm is still recorded, so downloads decompress as usual:

```rust
session.put(zstd_data).precompressed(Compression::Zstd).send().await?;
```

```python
session.put(zstd_data, precompressed="zstd")
```

To distinguish the two better, the put `compression` parameter is now deprecated and called `compress`. `Usecase` and multipart keep their `compression` parameter, which was already record-only.

Two behavior fixes fall out: batch classification no longer inflates precompressed part sizes by `zstd_safe::compress_bound`, so large precompressed parts stay batchable; and the Python client no longer records a meaningless uncompressed size for precompressed uploads.

Fix FS-394


lcian (approved):
Looks good!

PR #588, opened 2026-08-05
feat(cogs): Create inventory-tracker crate for storage oplog

matt-codecov:
Ref FS-210
Related to https://github.com/getsentry/sentry-kafka-schemas/pull/497

Creates the `inventory-tracker` crate in our project that we will use to emit Kafka messages for a change stream that we will turn into a COGS pipeline.

This crate isn't Objectstore-specific at all. If/when another service wants to use it, we can move it to a separate repository and set up a release process for it. I just couldn't be bothered if nobody's asking for it.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -21,6 +21,7 @@ debug = true
 lto = "thin"
 
 [workspace.dependencies]
+objectstore-inventory-tracker = { path = "objectstore-inventory-tracker" }
 objectstore-client = { path = "clients/rust" }
 objectstore-log = { path = "objectstore-log" }
 objectstore-metrics = { path = "objectstore-metrics" }
@@ -42,6 +43,7 @@ async-trait = "0.1.89"
 axum = "0.8.9"
 base64 = "0.22.1"
 bigtable_rs = { git = "https://github.com/getsentry/bigtable_rs.git", rev = "4cb75bc5e5f87204363973f6302107768e64972e" }
+blake3 = "1.8.5"
 bytes = "1.12.0"
 bytesize = "2.4.0"
 chrono = "0.4.45"
@@ -81,6 +83,10 @@ rustls = { version = "0.23.40", default-features = false }
 secrecy = "0.10.3"
 sentry = "0.48.3"
 sentry-options = "1.2.4"
+# - `ssl` is required for SASL/SCRAM
+# - `librdkafka` must be built from source either way, and `cmake-build` is the
+#    recommended way to do it
+rdkafka = { version = "0.39.0", features = ["cmake-build", "ssl", "tracing"] }
 serde = { version = "1.0.228", features = ["derive"] }
 serde_json = "1.0.150"
 serde_yaml = "0.9.34-deprecated"
--- a/objectstore-inventory-tracker/src/kafka.rs
+++ b/objectstore-inventory-tracker/src/kafka.rs
@@ -1,0 +1,242 @@
+//! A [`Producer`] backed by [`rdkafka`].
+
+use std::collections::HashMap;
+use std::time::Duration;
+
+use rdkafka::ClientConfig;
+use rdkafka::client::ClientContext;
+use rdkafka::error::KafkaError;
+use rdkafka::producer::{
+    BaseRecord, DeliveryResult, Producer as RdKafkaProducer, ProducerContext, ThreadedProducer,
+};
+
+use crate::producer::Producer;
+
+/// Connection settings for [`KafkaProducer`].
+#[derive(Clone, Debug, serde::Deserialize, serde::Serialize)]
+#[serde(default)]
+pub struct KafkaConfig {
+    /// Topic to produce to.
+    pub topic: String,
+
+    /// Broker addresses.
+    pub bootstrap_servers: Vec<String>,
+
+    /// Additional librdkafka properties, passed through verbatim.
+    ///
+    /// SASL credentials, compression, and buffering limits go here.
+    pub override_params: HashMap<String, String>,
+}
+
+impl Default for KafkaConfig {
+    fn default() -> Self {
+        Self {
+            topic: "shared-resources-inventory".to_owned(),
+            bootstrap_servers: Vec::new(),
+            override_params: HashMap::new(),
+        }
+    }
+}
+
+/// Something that went wrong producing to Kafka.
+#[derive(Debug, thiserror::Error)]
+pub enum KafkaProducerError {
+    /// The producer could not be created from the given configuration.
+    #[error("failed to create kafka producer")]
+    InvalidConfig(#[source] KafkaError),
+
+    /// The record could not be enqueued.
+    ///
+    /// This may mean the local queue is full, which may be caused by the broker being
+    /// unreachable or backed up.
+    #[error("failed to enqueue inventory record")]
+    SendFailed(#[source] KafkaError),
+
+    /// The local queue was not emptied before the flush timeout elapsed.
+    #[error("failed to flush inventory records")]
+    FlushFailed(#[source] KafkaError),
+}
+
+/// Called for each record that fails to deliver.
+///
+/// Invoked from a librdkafka background thread, so it must not block.
+pub type OnDeliveryFailure = Box<dyn Fn(&KafkaError) + Send + Sync>;
+
+/// Reports delivery outcomes.
+///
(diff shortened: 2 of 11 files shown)


Cargo.toml
```diff
@@ -1,6 +1,12 @@
 [workspace]
 resolver = "3"
-members = ["objectstore-*", "clients/rust", "stresstest", "bigtable-bench"]
+members = [
+    "objectstore-*",
+    "inventory-tracker",
```
jan-auer: As per convention, all crates that belong to objectstore server are to be prefixed with `objectstore-` so they are captured here automatically.

If we move this out of objectstore as a general crate, we can drop the prefix. Existing examples for this are the `objectstore-typed-options-*` crates.

Cargo.toml
```diff
@@ -42,6 +49,7 @@ async-trait = "0.1.89"
 axum = "0.8.9"
 base64 = "0.22.1"
 bigtable_rs = { git = "https://github.com/getsentry/bigtable_rs.git", rev = "4cb75bc5e5f87204363973f6302107768e64972e" }
+blake3 = "1.5"
```
jan-auer: Please use full version number

objectstore-inventory-tracker/src/lib.rs
```diff
@@ -0,0 +1,94 @@
...
+
+#![warn(missing_docs)]
+#![warn(missing_debug_implementations)]
+
+mod producer;
+mod record;
+mod tracker;
+
+#[cfg(feature = "kafka")]
+mod kafka;
```
jan-auer: As an idea, what if we make this a `pub mod` instead and always import with the module stated? This is a common pattern where there are many different implementations of something with their own errors and utility types. 

Same for the test utils below. 

FTR, it makes sense to keep the inventory tracker etc exposed at top level.

objectstore-inventory-tracker/src/producer.rs
```diff
@@ -0,0 +1,100 @@
...
+/// There is no `flush()`. `sentry_arroyo` doesn't provide a way to flush on demand so it
+/// would only be a no-op until other transport backends are introduced.
+pub trait Producer {
+    /// What can go wrong when sending.
+    type Error;
+
+    /// Enqueues one record.
+    ///
+    /// `key` controls which partition receives the message.
+    fn send(&self, key: &[u8], payload: Vec<u8>) -> Result<(), Self::Error>;
```
jan-auer, line 24: At least in arroyo / librdkafka, my understanding is that because of message batching sending a message can also fail after they were initially recorded. With the fire-and-forget style `send` API, these errors will not be returned anymore. 

For us, this is likely fine - we want non-blocking best effort submission. Though we should pick one of the below options:
 - Acknowledge this in the doc comment and call out that not all errors can be captured
 - Make the signature infallible and move the responsibility for handling errors to the producer implementation.

inventory-tracker/src/kafka.rs
```diff
@@ -0,0 +1,146 @@
...
+            topic,
+            bootstrap_servers,
+            override_params,
+        } = config;
+
+        let arroyo_config =
+            ArroyoKafkaConfig::new_producer_config(bootstrap_servers, Some(override_params));
+
+        let producer =
+            catch_unwind(AssertUnwindSafe(|| ArroyoKafkaProducer::new(arroyo_config)))
```
jan-auer: I think this creates a threaded producer. We will have to look into shutdown behavior, since dropping that producer will simply drop the poll thread and might not produce all messages that are currently in queue. 

If that is the case, we will have to either expose or build a mechanism that suspends until everything is out. Objectstore has ways for backends to delay shutdown, but we need to hook this up here.

jan-auer (commented):
Initial review. Some of the below comments can be left for follow-up to get a first version in, as long as they don't change the overall concept and can be fixed in isolation. Particularly, shutdown is a larger topic.

inventory-tracker/src/kafka.rs
jan-auer: There's a `flush` method exposed now, which should be enough.

jan-auer (approved):
Looks good. See the open comments before merge, please.

Either in this PR or in a follow up, I'd suggest to add metrics, especially for errors.

objectstore-inventory-tracker/src/producer.rs
matt-codecov: The `on_delivery_failure` argument to the Kafka producer should give callers a hook into failed messages. Failures here are when, like, the local queue is full or something.

jan-auer: This is the high-level interface, so callers should not have a concept about Kafka and should not need to hook the delivery callback. This is something the kafka producer impl can do internally.

To clarify my original message: we don't have to change anything, but we can write this into the doc comment so callers know that `Ok()` is not a guarantee.

objectstore-inventory-tracker/src/kafka.rs
```diff
@@ -0,0 +1,225 @@
...
+}
+
+impl ClientContext for DeliveryReporter {}
+
+impl ProducerContext for DeliveryReporter {
+    type DeliveryOpaque = ();
+
+    fn delivery(&self, result: &DeliveryResult<'_>, _opaque: Self::DeliveryOpaque) {
+        if let Err((error, _)) = result {
+            tracing::warn!(%error, "failed to deliver inventory record");
```
jan-auer: There's dedicated syntax to pass errors so they get captured properly. 

```suggestion
            tracing::warn!(!!error, "failed to deliver inventory record");
```

objectstore-inventory-tracker/src/producer.rs
```diff
@@ -0,0 +1,148 @@
...
+
+    /// Blocks until enqueued records have been delivered, or `timeout` elapses.
+    ///
+    /// Sending is asynchronous, so records handed over just before a process exits are
+    /// still sitting in a local queue. Call this during shutdown to deliver them.
+    fn flush(&self, timeout: Duration) -> Result<(), Self::Error>;
```
jan-auer: Note that backends use `join` terminology for this, but `flush` can also work here. 

Let's make this `async` on the interface, so we can join it in the same way that we join backends. Internally, try to delegate to a `spawn_blocking` so that we do not block the executor when flushing the producer.

objectstore-inventory-tracker/src/producer.rs
```diff
@@ -0,0 +1,148 @@
...
+
+/// Erases `producer` into a [`SharedProducer`].
+///
+/// ```
+/// use objectstore_inventory_tracker::{InventoryTracker, NoopProducer, shared_producer};
+///
+/// let producer = shared_producer(NoopProducer);
+/// let tracker = InventoryTracker::new(producer.clone(), "my_gcs_bucket", 1.0);
+/// ```
+pub fn shared_producer<P>(producer: P) -> SharedProducer
```
jan-auer: I think you can make this a provided method on the `Producer` trait so that you simply have to call `.shared()` for this.

objectstore-inventory-tracker/src/tracker.rs
```diff
@@ -0,0 +1,490 @@
...
+    #[allow(clippy::too_many_arguments)]
+    pub fn write(
+        &self,
+        storage_key: &str,
+        app_feature: &str,
+        size: u64,
+        timestamp: SystemTime,
+        expiration_time: Option<SystemTime>,
+        organization_id: Option<u64>,
+        project_id: Option<u64>,
```
jan-auer, line 155: Not blocking for this PR, but for later as we integrate and revise the API of this layer: Here we may want to take in `scopes` and `usecase` and then map this internally. This way, the backends and whoever is using inventory tracker don't need to perform this mapping.

jan-auer (approved):
Thanks! Approving for merge, provided the comments below are solved

objectstore-inventory-tracker/src/kafka.rs
matt-codecov: note to self, look into objectstore's tracing wrapper macro and basically inline what the `!!` does

objectstore-inventory-tracker/src/tracker.rs
matt-codecov: inventory-tracker is meant to be service-agnostic so we can extract it into a separate crate if another sentry service needs it. so it shouldn't use these objectstore concepts in its API

jan-auer (approved):

PR #590, opened 2026-08-06
feat(cogs): ChangeStream trait, preparation for backend instrumentation

matt-codecov:
Ref FS-210
Depends on https://github.com/getsentry/objectstore/pull/588

- `ChangeStream` trait that `Backend` impls will use to record write/update/delete operations.
- `CostTrackerStream` impl and config to publish change stream events to Kafka through `inventory-tracker`
- `NoopStream` stub impl is used by default

While `ChangeStream` is abstract and we can implement a `MultiStream` wrapper that dispatches changes to multiple sinks, currently the abstraction is only in the code and not in the config. `CostTrackerStream` is configured through `CostTrackerConfig` (service-wide, Kafka details, used in later PR) and `CostTrackerStreamConfig` (on each `Backend` impl, added here). It's assumed that other `ChangeStream` implementations will similarly be named directly in config.

Backends are not instrumented yet and `objectstore-server` doesn't expose configuring Kafka yet.

### DIFF
--- a/Dockerfile.cross
+++ b/Dockerfile.cross
@@ -4,7 +4,7 @@ FROM rust:slim-bookworm
 RUN dpkg --add-architecture amd64 \
     && apt-get update -qq \
     && apt-get upgrade -y \
-    && apt-get install -y --no-install-recommends make protobuf-compiler libprotobuf-dev pkg-config git libssl-dev:amd64 gcc-x86-64-linux-gnu g++-x86-64-linux-gnu \
+    && apt-get install -y --no-install-recommends make protobuf-compiler libprotobuf-dev pkg-config git libssl-dev:amd64 libcurl4-openssl-dev:amd64 gcc-x86-64-linux-gnu g++-x86-64-linux-gnu \
     && rm -rf /var/lib/apt/lists/*
 
 RUN rustup target add x86_64-unknown-linux-gnu
--- a/objectstore-server/src/state.rs
+++ b/objectstore-server/src/state.rs
@@ -9,6 +9,7 @@ use std::time::Duration;
 use anyhow::Result;
 use bytes::Bytes;
 use futures_util::Stream;
+use objectstore_service::change_stream::ChangeStreamFactory;
 use objectstore_service::concurrency::ConcurrencyLimiter;
 use objectstore_service::id::ObjectContext;
 use objectstore_service::{StorageService, backend};
@@ -61,7 +62,8 @@ impl Services {
         #[cfg(target_os = "linux")]
         tokio::spawn(track_allocator_metrics(config.runtime.metrics_interval));
 
-        let backend = backend::from_config(config.storage.clone()).await?;
+        let backend =
+            backend::from_config(config.storage.clone(), &ChangeStreamFactory::default()).await?;
         let concurrency = ConcurrencyLimiter::new(config.service.max_concurrency)
             .with_queue(config.service.concurrency_queue)
             .with_timeout(config.service.concurrency_timeout)
--- a/objectstore-service/Cargo.toml
+++ b/objectstore-service/Cargo.toml
@@ -20,6 +20,7 @@ futures-util = { workspace = true }
 gcp_auth = { workspace = true }
 humantime = { workspace = true }
 humantime-serde = { workspace = true }
+objectstore-inventory-tracker = { workspace = true, features = ["kafka"], optional = true }
 objectstore-log = { workspace = true }
 objectstore-metrics = { workspace = true }
 objectstore-types = { workspace = true }
@@ -36,8 +37,16 @@ tonic = { workspace = true }
 tracing = { workspace = true }
 uuid = { workspace = true, features = ["v7"] }
 
+[features]
+# Publishes a per-record change stream used to attribute storage cost to app features.
+# Off by default: it pulls in librdkafka, whose native build needs a toolchain the
+# cross-build image would otherwise not need, and sandbox deployments have no topic or
+# consumer for the records anyway.
+storage-cogs = ["dep:objectstore-inventory-tracker"]
+
 [dev-dependencies]
 futures = { workspace = true }
+objectstore-inventory-tracker = { workspace = true, features = ["test-utils"] }
 tempfile = { workspace = true }
 tokio = { workspace = true, features = ["test-util"] }
 zstd = { workspace = true }
--- a/objectstore-service/src/backend/bigtable.rs
+++ b/objectstore-service/src/backend/bigtable.rs
@@ -47,7 +47,9 @@ use crate::backend::common::{
     Backend, DeleteResponse, GetResponse, HighVolumeBackend, MetadataResponse, PutResponse,
     TieredGet, TieredMetadata, TieredWrite, Tombstone,
 };
-use crate::change_stream::CostTrackerStreamConfig;
+use crate::change_stream::{
+    ChangeStream, ChangeStreamFactory, CostTrackerStreamConfig, flush_change_stream,
+};
 use crate::error::{Error, Result};
 use crate::gcp_auth::PrefetchingTokenProvider;
 use crate::id::ObjectId;
@@ -184,6 +186,8 @@ pub struct BigTableBackend {
     instance_path: String,
     table_path: String,
     table_name: String,
+
+    change_stream: Arc<dyn ChangeStream>,
 }
 
 impl fmt::Debug for BigTableBackend {
@@ -708,15 +712,19 @@ impl BigTableBackend {
     ///
     /// Pass an `endpoint` in the config to connect to a local emulator; omit it to use real GCP
     /// credentials. `connections` controls the gRPC connection pool size (defaults to 1).
-    pub async fn new(config: BigTableConfig) -> anyhow::Result<Self> {
+    pub async fn new(
+        config: BigTableConfig,
+        streams: &ChangeStreamFactory,
+    ) -> anyhow::Result<Self> {
         let BigTableConfig {
             endpoint,
             project_id,
             instance_name,
             table_name,
             connections,
-            cogs: _,
+            cogs,
         } = config;
+        let change_stream = streams.build(cogs.as_ref());
 
         let bigtable = if let Some(ref endpoint) = endpoint {
             BigTableConnection::new_with_emulator(
@@ -749,6 +757,7 @@ impl BigTableBackend {
             instance_path: format!("projects/{project_id}/instances/{instance_name}"),
             table_path: client.get_full_table_name(&table_name),
             table_name,
+            change_stream,
         })
     }
 
@@ -966,6 +975,10 @@ impl Backend for BigTableBackend {
         self.mutate(path, [delete_row_mutation()], "delete").await?;
 
         Ok(())
+    }
+
+    async fn join(&self) {
+        flush_change_stream(&self.change_stream).await;
     }
 }
 
@@ -1326,16 +1339,19 @@ mod tests {
     //
     // Refer to the readme for how to set up the emulator.
 
-    async fn create_test_backend() -> Result<BigTableBackend> {
-        BigTableBackend::new(BigTableConfig {
+    fn test_config() -> BigTableConfig {
+        BigTableConfig {
             endpoint: Some("localhost:8086".into()),
             project_id: "testing".into(),
             instance_name: "objectstore".into(),
             table_name: "objectstore".into(),
             connections: None,
             cogs: None,
-        })
-        .await
+        }
+    }
+
+    async fn create_test_backend() -> Result<BigTableBackend> {
+        BigTableBackend::new(test_config(), &ChangeStreamFactory::default()).await
     }
 
     fn make_id() -> ObjectId {
--- a/objectstore-service/src/backend/bigtable.rs
+++ b/objectstore-service/src/backend/bigtable.rs
@@ -47,6 +47,7 @@ use crate::backend::common::{
     Backend, DeleteResponse, GetResponse, HighVolumeBackend, MetadataResponse, PutResponse,
     TieredGet, TieredMetadata, TieredWrite, Tombstone,
 };
+use crate::change_stream::CostTrackerStreamConfig;
 use crate::error::{Error, Result};
 use crate::gcp_auth::PrefetchingTokenProvider;
 use crate::id::ObjectId;
@@ -124,6 +125,19 @@ pub struct BigTableConfig {
     ///
     /// - `OS__STORAGE__CONNECTIONS=16` (optional)
     pub connections: Option<usize>,
+
+    /// Reports what this backend stores, for per-usecase cost attribution.
+    ///
+    /// # Default
+    ///
+    /// `None`, which disables reporting for this backend.
+    ///
+    /// # Environment Variables
+    ///
+    /// - `OS__STORAGE__COGS__SHARED_RESOURCE_ID=bigtable_objectstore`
+    /// - `OS__STORAGE__COGS__SAMPLE_RATE=1.0` (optional)
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub cogs: Option<CostTrackerStreamConfig>,
 }
 
 /// Connection timeout used for the initial connection to Bigtable.
@@ -701,6 +715,7 @@ impl BigTableBackend {
             instance_name,
             table_name,
             connections,
+            cogs: _,
         } = config;
 
         let bigtable = if let Some(ref endpoint) = endpoint {
@@ -1318,6 +1333,7 @@ mod tests {
             instance_name: "objectstore".into(),
             table_name: "objectstore".into(),
             connections: None,
+            cogs: None,
         })
         .await
     }
--- a/objectstore-service/src/change_stream/cost_tracker.rs
+++ b/objectstore-service/src/change_stream/cost_tracker.rs
@@ -1,0 +1,265 @@
+//! The change stream implementation that reports through an [`InventoryTracker`].
+
+use std::fmt;
+use std::time::{Duration, SystemTime};
+
+use objectstore_inventory_tracker::{BoxError, InventoryTracker, Producer};
+
+use crate::change_stream::{
+    ChangeStream, CostTrackerStreamConfig, SCOPE_ORGANIZATION, SCOPE_PROJECT, scope_id,
+};
+use crate::id::ObjectId;
+
+/// Reports through an [`InventoryTracker`], which hashes each [`ObjectId`] both to
+/// anonymize it and to decide whether it is sampled. See [`objectstore_inventory_tracker`]
+/// for the record format.
+///
+/// Logs, counts, and swallows errors returned by the [`InventoryTracker`].
+pub struct CostTrackerStream<P: Producer> {
+    tracker: InventoryTracker<P>,
+}
+
+impl<P: Producer> CostTrackerStream<P> {
+    /// Reports changes through `producer`, as described by `config`.
+    pub fn new(producer: P, config: &CostTrackerStreamConfig) -> Self {
+        Self {
+            tracker: InventoryTracker::new(
+                producer,
+                &config.shared_resource_id,
+                config.sample_rate,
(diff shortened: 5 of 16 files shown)


objectstore-service/src/backend/common.rs
```diff
@@ -29,6 +32,28 @@ pub type MetadataResponse = Option<Metadata>;
 /// Backend response for delete operations.
 pub type DeleteResponse = ();
 
+/// Wraps a stream to count the total bytes yielded by successful chunks.
+///
+/// Returns the shared counter and the wrapped stream. The counter is incremented
+/// as the stream is consumed, so read it only after the stream is exhausted.
+pub(crate) fn counting_stream<S, E>(
```
jan-auer: This is moved but not used anywhere other than the previous usage site?

objectstore-service/src/backend/inventory.rs
```diff
@@ -0,0 +1,303 @@
...
+/// Scope key holding the Sentry organization ID.
+const SCOPE_ORGANIZATION: &str = "org";
+/// Scope key holding the Sentry project ID.
+const SCOPE_PROJECT: &str = "project";
+
+/// Wrapper trait for [`objectstore_inventory_tracker::InventoryTracker`]. Used to publish
+/// a change stream for our storage backends.
+///
+/// See [module docs](self).
+pub trait InventorySink: fmt::Debug + Send + Sync + 'static {
```
jan-auer: Calling out that this interface is close to what is being created in #582. I don't have a concrete design in mind yet, but it seems the "inventory" concept would be capable of solving both problems:
 - Storage cost tracking, _which requires to know which files are stored until when_
 - GC / eviction, _which requires to know which files are stored until when_

aldy505: Ah yes. Agreed with this.

jan-auer: Apart from that we don't want to couple the two in-flight PRs too much, so this is probably more relevant for the other PR:

The biggest difference is how they are needed: Cost tracking requires the inventory _independent_ of the backend, and will export it via kafka to a remote system. Conversely, the S3 and FS backends need the inventory as part of their inner workings. Per design, GC is part of the backend's responsibility. 

We might be able to separate this cleanly if we:
 - Hook the inventory up in the service like the PR here does (sort of like an event bus)
 - Allow multiple inventory exporters
 - One is the kafka sink used for cost tracking
 - Another one is a database sink (like sqlite, PG) for GC, configured/hooked when S3/FS are used
    - GC is still an external command then, that queries said database

objectstore-service/src/backend/common.rs
matt-codecov: belongs in next PR probably. gcs needs to use it

objectstore-service/src/backend/inventory.rs
matt-codecov: my direction here is very much "fail open" and errors are logged and swallowed. if inventory tracking is load-bearing for GC in some backends, that'll need to change

correct me if i'm wrong: for self-hosted, in order to support the automatic TTI/TTL GC that we advertise, we have to implement it ourselves atop filesystem and S3 as we anticipate those will be popular in self-hosted deployments?

objectstore-service/src/backend/common.rs
matt-codecov: moved to next PR

objectstore-service/src/backend/change_stream.rs
```diff
@@ -0,0 +1,303 @@
...
+/// Scope key holding the Sentry organization ID.
+const SCOPE_ORGANIZATION: &str = "org";
+/// Scope key holding the Sentry project ID.
+const SCOPE_PROJECT: &str = "project";
+
+/// Wrapper trait for [`objectstore_inventory_tracker::InventoryTracker`]. Used to publish
+/// a change stream for our storage backends.
+///
+/// See [module docs](self).
+pub trait ChangeStream: fmt::Debug + Send + Sync + 'static {
```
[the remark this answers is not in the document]
matt-codecov: oops lol artifact of commit splitting

objectstore-service/src/backend/inventory.rs
jan-auer: That's right, and that's what @aldy505's PR linked in my initial comment does. The trait is called `Keeper` there and it's intended to be integrated directly into the backends. Conceptually, they hold similar responsibilities.

While for cogs accounting errors are less critical, we should still find an implementation that's correct by design or has well-understood failure modes. Failing loudly/openly does work, but we need to look closely at what happens during concurrent requests.

objectstore-server/src/state.rs
```diff
@@ -61,7 +62,8 @@ impl Services {
         #[cfg(target_os = "linux")]
         tokio::spawn(track_allocator_metrics(config.runtime.metrics_interval));
 
-        let backend = backend::from_config(config.storage.clone()).await?;
+        let backend =
+            backend::from_config(config.storage.clone(), &ChangeStreamFactory::default()).await?;
```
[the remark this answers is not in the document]
matt-codecov: future commit plugs in real config

Dockerfile.cross
```diff
@@ -4,7 +4,7 @@ FROM rust:slim-bookworm
 RUN dpkg --add-architecture amd64 \
     && apt-get update -qq \
     && apt-get upgrade -y \
-    && apt-get install -y --no-install-recommends make protobuf-compiler libprotobuf-dev pkg-config git libssl-dev:amd64 gcc-x86-64-linux-gnu g++-x86-64-linux-gnu \
+    && apt-get install -y --no-install-recommends make protobuf-compiler libprotobuf-dev pkg-config git libssl-dev:amd64 libcurl4-openssl-dev:amd64 gcc-x86-64-linux-gnu g++-x86-64-linux-gnu \
```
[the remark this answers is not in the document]
jan-auer: It would be good if we could disable this via a feature on the objectstore-server crate. That way, we won't have to build with rdkafka when we publish, nor does this have to be built for sandbox.

Edit: see my other comment.

objectstore-service/Cargo.toml
```diff
@@ -20,6 +20,7 @@ futures-util = { workspace = true }
 gcp_auth = { workspace = true }
 humantime = { workspace = true }
 humantime-serde = { workspace = true }
+objectstore-inventory-tracker = { workspace = true, features = ["kafka"] }
```
jan-auer: Let's make this `optional = true` and enable it with a dedicated feature flag. That way, we can control for individual builds, whether we have kafka-based inventory tracking included or not. For sandbox cross builds and for self-hosted releases in the future, we'll not include this. 

This also means that we do not have to add cmake to the cross-build Dockerfile anymore.

objectstore-service/src/change_stream/mod.rs
```diff
@@ -0,0 +1,127 @@
...
+    fn update(&self, _id: &ObjectId, _expires_at: Option<SystemTime>) {}
+
+    fn delete(&self, _id: &ObjectId) {}
+
+    fn flush(&self, _timeout: Duration) {}
+}
+
+/// Reads a scope value off `id` as an integer, if present and well-formed.
+fn scope_id(id: &ObjectId, scope: &str) -> Option<u64> {
+    id.scopes().get_value(scope)?.parse().ok()
```
jan-auer, line 123: Does usage-accountant require these to be numeric? Especially at this level we should not make assumptions about the type and expose org and project as strings, and let the consumer handle this.

If this is a requirement, let's do the conversion at the innermost level possible (in the kafka implementation, likely).

objectstore-service/src/change_stream/kafka.rs
```diff
@@ -0,0 +1,334 @@
...
+}
+
+/// A [`ChangeStream`] listener that reports onto a Kafka topic.
+///
+/// Reports through an [`InventoryTracker`], which hashes each [`ObjectId`] both to
+/// anonymize it and to decide whether it is sampled. See [`objectstore_inventory_tracker`]
+/// for the record format.
+///
+/// Logs, counts, and swallows errors returned by the [`InventoryTracker`].
+pub struct KafkaStream {
```
jan-auer: Let's rename this type. Technically, this can be any kind of InventoryTracker, it purely depends on which producer is used to instantiate it. 

It just so happens that `build_producer` and thus the change stream factory currently only have a kafka field; see my overall review comment for this.

objectstore-service/src/change_stream/kafka.rs
```diff
@@ -0,0 +1,334 @@
...
+    1.0
+}
+
+/// Configuration of the Kafka producer carrying every backend's Kafka listener.
+///
+/// Service-wide: one producer serves every backend that configures a
+/// [`ListenerConfig`], so connections and the send queue are not duplicated per backend.
+#[derive(Debug, Clone, Deserialize, Serialize)]
+#[serde(default)]
+pub struct SinkConfig {
```
jan-auer: This is a complete clone of `KafkaConfig`. We should probably make this an enum that allows for different kinds of InventoryTracker producers, one of which is the kafka variant. That way, the interface/config becomes extensible.

Even better: Don't assume any of that and just pass a complete `InventoryTracker` into the constructor of this type. That way, the concerns are split entirely.

jan-auer (changes requested):
Thanks, the documentation and tests made this easy to review.

There is a good foundation for separation of concerns, but the current config mixes them again:
- `ChangeStream` tracks logical changes to objects agnostic of how this information is interpreted. This is what backends are instrumented with, and it can be hooked in various ways.
- `InventoryTracker` is the component that maps this to cost-tracking semantics. 

There are several downsides: ChangeStream's config now needs to know about the possible implementations of inventory tracker, comes with an optional kafka config, there could now be different change streams per backend, it needs to know about shared resource IDs, and about which scopes are important for cost tracking. This makes it harder to use the change stream for GC. 

This can be solved:
 - Keep `ChangeStream` trait tied to objectstore-service concepts like it is now.
 - Instantiate ChangeStream globally as backends are being created, and pass it generically into the backends. This removes the need for an internal factory. 
 - Have every backend annotate its stream items with an identifier. Initially that can be hard coded, but that's the one thing we can make overridable in backend config (i.e. `gcs1` vs `gcs2`). 
 - The config for _which_ ChangeStream to instantiate should live adjacent to the backends. There's only a single one available now, the inventory tracker, and then we directly surface the inventory tracker's possible implementations there. 
- Inventory tracker and its config is the level at which we
  - map the built-in backend IDs to `shared_resource_id` (cost concern)
  - extract org ID and project ID from scopes (if available)
  - optionally map use case to app features (currently part of the usecase config)

objectstore-service/src/change_stream/mod.rs
matt-codecov: `usage-accountant` doesn't have proj/org IDs

the new kafka topic (which i just merged and released, oops) has them as integers. not too late to change, would just need a new PR and release

objectstore-service/src/change_stream/mod.rs
```diff
@@ -0,0 +1,124 @@
...
+/// Publishes the changes a single backend makes to the objects it stores.
+///
+/// See [module docs](self).
+#[async_trait::async_trait]
+pub trait ChangeStream: fmt::Debug + Send + Sync + 'static {
+    /// Reports that `id` now occupies `size` bytes. Used for new writes and overwrites.
+    fn write(&self, id: &ObjectId, size: u64, expires_at: Option<SystemTime>);
+
+    /// Reports that `id`'s expiration moved, with its stored size unchanged.
+    fn update(&self, id: &ObjectId, expires_at: Option<SystemTime>);
```
aldy505, line 82: I would prefer to replace `expires_at`, which uses `Option<SystemTime>`, with `ExpirationPolicy`. This way, if we want to implement a `ChangeStream` trait that has GC built in for S3/filesystem, I can know what kind of expiration policy the object has.

matt-codecov: why would GC need to know more than just the time that the object expires at? does it matter whether it's 30-day TTL vs 90-day TTL? (@jan-auer is thinking of removing TTI policies)

one thing that my current stack of PRs doesn't achieve but should: the change stream should use literally the same timestamp that is used to set when the object is supposed to expire. currently in a couple places I call `SystemTime::now()` and re-compute the expiration timestamp because the current code doesn't make the timestamp it used available to me. if this API took the `ExpirationPolicy` instead of the actual timestamp, i think we would _have_ to recompute the expiration time.

aldy505: Oh I didn't know about Jan's plan of removing TTI policies. Previously with TTI policies, I need to know how much I need to add to the expiration, as I can only get the object id, not the `expires_at` (as this is handled by GCS's policy, not something objectstore manage initially).

jan-auer: > with TTI policies, I need to know how much I need to add to the expiration

While TTI is still around, the change stream currently outputs an `update` record with the new expiry time on TTI bump too. This is nice, because GC can just rely on that update and doesn't have to deal with expiration policies itself. 

> the change stream should use literally the same timestamp that is used to set when the object is supposed to expire

I'm working on a change that pulls TTI bumps out of the backends into the service layer (needed to remove TTI). With this, there will be one authoritative `expires_at` that we can use for the change stream, too. I'd suggest we wait for this change to land, as it makes things a lot simpler.

objectstore-service/Cargo.toml
```diff
@@ -36,8 +37,16 @@ tonic = { workspace = true }
 tracing = { workspace = true }
 uuid = { workspace = true, features = ["v7"] }
 
+[features]
+# Publishes a per-record change stream used to attribute storage cost to app features.
+# Off by default: it pulls in librdkafka, whose native build needs a toolchain the
+# cross-build image would otherwise not need, and sandbox deployments have no topic or
+# consumer for the records anyway.
```
jan-auer, line 44: Let's put this into docs (e.g. crate-level docs) rather than cargo.toml, so that this is also rendered on [deployed docs](https://getsentry.github.io/objectstore/rust/objectstore_service/index.html).

objectstore-service/src/change_stream/cost_tracker.rs
```diff
@@ -0,0 +1,263 @@
+//! The change stream implementation that reports through an [`InventoryTracker`].
+
+use std::fmt;
+use std::time::{Duration, SystemTime};
+
+use objectstore_inventory_tracker::{BoxError, InventoryTracker, Producer};
+
+use super::{ChangeStream, CostTrackerStreamConfig, SCOPE_ORGANIZATION, SCOPE_PROJECT, scope_id};
```
jan-auer: Other than for test modules, prefer to use `crate::` paths.

objectstore-service/src/change_stream/mod.rs
```diff
@@ -0,0 +1,124 @@
...
+    pub shared_resource_id: String,
+
+    /// Proportion of records to report, in `[0, 1]`.
+    ///
+    /// `1.0` reports every change. It can be lowered if the stream is under too much load
+    /// but beware: when the sample rate decreases, records that used to be tracked will
+    /// no longer be tracked. Stream consumers may have inconsistent state for them until
+    /// they expire.
+    #[serde(default = "default_sample_rate")]
+    pub sample_rate: f64,
```
jan-auer, line 65: I would consider this a global property rather than per-change stream. Is this here because we expect backends to have vastly different throughput in tiered setups?

objectstore-service/src/backend/bigtable.rs
```diff
@@ -124,6 +127,19 @@ pub struct BigTableConfig {
...
+    /// # Default
+    ///
+    /// `None`, which disables reporting for this backend.
+    ///
+    /// # Environment Variables
+    ///
+    /// - `OS__STORAGE__STORAGE_COGS__SHARED_RESOURCE_ID=bigtable_objectstore`
+    /// - `OS__STORAGE__STORAGE_COGS__SAMPLE_RATE=1.0` (optional)
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub storage_cogs: Option<CostTrackerStreamConfig>,
```
jan-auer: _nit_: On the backends we can call it just `cogs` - given that this is the configuration for a storage backend, it can be implied that this is talking about storage. 

Structure in yaml and the env var show this nicely.

objectstore-service/src/backend/bigtable.rs
```diff
@@ -170,6 +186,10 @@ pub struct BigTableBackend {
     instance_path: String,
     table_path: String,
     table_name: String,
+
+    /// Records each write/update/delete performed by the backend. May be
+    /// [`NoopStream`](crate::change_stream::NoopStream).
+    change_stream: Arc<dyn ChangeStream>,
```
jan-auer:
```suggestion
    change_stream: Arc<dyn ChangeStream>,
```

objectstore-service/src/service.rs
```diff
@@ -465,19 +469,31 @@ mod tests {
...
             bucket: "test-bucket".into(),
+            storage_cogs: None,
         };
 
-        let high_volume = Box::new(BigTableBackend::new(bigtable_config).await.unwrap());
-        let long_term = Box::new(GcsBackend::new(gcs_config.clone()).await.unwrap());
+        let high_volume = Box::new(
+            BigTableBackend::new(bigtable_config, &ChangeStreamFactory::default())
+                .await
+                .unwrap(),
```
jan-auer, line 483: It's already here, but to make tests a bit more ergonomic, I could think of two improved APIs for this:

1. A test-only `ChangeStreamFactory::empty()` that synchronously creates a non-kafka factory. That way, no awat and no unwrap.
2. Configure the factory via a builder method (`BigTableBackend::new(...).change_stream(factory)`). This way, we can leave it out in test code and only set it in prod. Only disadvantage is that it would now be structurally possible to forget supplying one in production code.

objectstore-service/src/change_stream/factory.rs
```diff
@@ -0,0 +1,108 @@
...
+        let Some(config) = config else {
+            return Arc::new(NoopStream);
+        };
+
+        let Some(producer) = self.producer.clone() else {
+            objectstore_log::error!(
+                shared_resource_id = config.shared_resource_id,
+                "backend configured a change stream but the service has no usable \
+                 transport; this backend will report nothing"
+            );
```
jan-auer: Please double-check if this is the behavior we actually want. This means we can no longer have "default" config that sets shared resource IDs for our backends while the global cost tracking is unconfigured. 

I'd prefer we return `NoopStream` silently here - the factory is still instantiating correctly what it should. 

We could consider adding an error to the _other_ branch, however. If cost tracking is configured globally but a backend doesn't supply its config, it now doesn't provide expected cost data.

jan-auer (approved):
Looks good, thanks! See minor comments below.

objectstore-service/src/change_stream/mod.rs
matt-codecov: each `InventoryTracker` can have its own sample rate, and each backend gets its own `InventoryTracker`, so it's natural enough to allow a separate rate per backend. and sampling makes a backend's data less powerful (small orgs/projects will likely be omitted which slants org/project analysis towards bigger orgs) so we want to target sampling to only the backends that are stressing the Kafka topic

we shouldn't need any sampling for a while, but if usage grows with our current Bigtable/GCS breakdown we'd want to sample Bigtable before we sample GCS

objectstore-service/src/service.rs
matt-codecov: maybe i'm misunderstanding. the `.await.unwrap()` aren't from the change stream factory, they were already there. `ChangeStreamFactory::default()` is synchronous, non-erroring, and returns a non-kafka factory.

objectstore-service/src/change_stream/factory.rs
matt-codecov: i'll log a warning in both cases. neither are P0 INC-causing issues, but configuring one side without the other is not meaningful and if you configure one you surely mean to follow up and configure the other.

not hugely important, i can follow up in an other PR if you feel strongly

PR #592, opened 2026-08-06
fix(types): Restrict the duration wire format to days and smaller

jan-auer:
Durations in the `x-sn-expiration` header are now emitted with only the `d`, `h`, `m`, and `s` units. Anything longer than a day stays in days, so a 400-day TTL serializes as `400d`, never `1year 1month 5days`. The new `objectstore_types::duration` module owns the format and documents it explicitly, including that seconds are the smallest unit and sub-second remainders are truncated.

This fixes a real round-trip bug. The Python client only understood `w/d/h/m/s`, so when the server formatted a TTL above a year it dropped the `1year` component and read `1month` as one minute — any expiration over a year was reported wrong by `get` and `head`. Uploads were unaffected.

Parsing stays lenient and still accepts weeks, months, and years, because the GCS backend persists the formatted string as object metadata and older objects carry those units. That leniency is deliberately not part of the documented format, so clients do not have to reproduce it: the Python client now formats exactly like the server and rejects units outside the contract instead of silently mis-parsing them.

Server-side use-case validation errors use the same formatter, so the durations in those messages match what clients see on the wire.

Fixes FS-433

### DIFF
--- a/clients/python/tests/test_metadata.py
+++ b/clients/python/tests/test_metadata.py
@@ -1,0 +1,56 @@
+from datetime import timedelta
+
+import pytest
+from objectstore_client.metadata import (
+    TimeToIdle,
+    TimeToLive,
+    format_expiration,
+    format_timedelta,
+    parse_expiration,
+    parse_timedelta,
+)
+
+# Durations and their wire format, used in both directions.
+ROUND_TRIP_CASES = [
+    (timedelta(0), "0s"),
+    (timedelta(seconds=30), "30s"),
+    (timedelta(minutes=1), "1m"),
+    (timedelta(hours=1), "1h"),
+    (timedelta(days=1), "1d"),
+    (timedelta(days=2, hours=3, seconds=4), "2d 3h 4s"),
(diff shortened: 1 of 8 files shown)


clients/python/tests/test_metadata.py
```diff
@@ -0,0 +1,71 @@
...
+
+
+@pytest.mark.parametrize(("delta", "formatted"), ROUND_TRIP_CASES)
+def test_parse_timedelta(delta: timedelta, formatted: str) -> None:
+    assert parse_timedelta(formatted) == delta
+
+
+def test_format_timedelta_truncates_sub_second_remainder() -> None:
+    assert format_timedelta(timedelta(milliseconds=1500)) == "1s"
+    assert format_timedelta(timedelta(milliseconds=500)) == "0s"
```
jan-auer, line 39: The sub-second truncation is fine, though we can think about rejecting such short lifetimes entirely.

lcian (approved):
Looks good to me.

PR #593, opened 2026-08-06
feat(server): Add Resumable Upload API

jan-auer:
Implements the Resumable Upload interface on the backend trait and the API endpoints.
By default, all backends decline with `501 Not Implemented`.

[Spec](https://app.notion.com/p/sentry/Project-Resumable-Uploads-3b38b10e4b5d8094a7eccc5dcae7ecc2)

Close FS-470


lcian:
bugbot run
@sentry review

lcian:
bugbot run
@sentry review

lcian:
bugbot run
@sentry review

lcian:
bugbot run
@sentry review

lcian:
bugbot run
@sentry review

objectstore-service/src/backend/common.rs
```diff
@@ -72,6 +73,56 @@ pub trait Backend: fmt::Debug + Send + Sync + 'static {
     fn as_multipart_upload_backend(&self) -> Result<&dyn MultipartUploadBackend> {
         Err(Error::NotImplemented)
     }
+
+    /// Creates a resumable upload session for the object at `id`.
+    ///
+    /// Object metadata and its total length are declared upfront and cannot be mutated
+    /// during the upload.
+    ///
+    /// Returns [`Error::NotImplemented`] when the backend refuses to create a session.
```
lcian: I looked hard for a better status code to distinguish "backend refuses it for this particular object" vs "backend doesn't support it at all", but it seems there's no suitable status code, so for now I'm using 501 to represent both states.

lcian:
bugbot run
@sentry review

objectstore-server/src/auth/service.rs
```diff
@@ -186,4 +187,57 @@ impl AuthAwareService {
             .complete_multipart(id, upload_id, parts)
             .await?)
     }
+
+    // --- Resumable upload operations ---
+    //
+    // Every operation requires `ObjectWrite`, including the two that do not obviously write:
+    // an offset query can commit an assembled object, and canceling a session discards an
+    // in-progress upload rather than deleting an object. So `DELETE ?session=` needs write
```
jan-auer: Can we move these down into the methods so it's closer to the permission? This way it is easier to find the commend when reading the method implementation.

objectstore-server/src/endpoints/resumable.rs
```diff
@@ -0,0 +1,481 @@
...
+    /// An existing session to continue or cancel.
+    ExistingSession,
+}
+
+impl ResumableQuery {
+    /// Classifies a request that may create a session or act on one.
+    ///
+    /// # Errors
+    ///
+    /// Returns [`ApiError::Client`] if both parameters are present.
```
jan-auer: What if we're lenient in this case and ignore the `upload_type` parameter? Technically it is not incorrect.

objectstore-server/src/endpoints/resumable.rs
```diff
@@ -0,0 +1,481 @@
...
+/// Which resumable session a request on an object route targets.
+#[derive(Debug)]
+pub(super) enum ResumableTarget {
+    /// A new session to create for the object addressed by the request.
+    NewSession,
+    /// An existing session to continue or cancel.
+    ExistingSession,
+}
+
+impl ResumableQuery {
```
jan-auer: _nit_: Move impl directly after struct declaration.

objectstore-server/src/endpoints/resumable.rs
```diff
@@ -0,0 +1,481 @@
...
+    Resumable,
+}
+
+/// The resumable protocol's query parameters, as seen on a regular object route.
+#[derive(Debug, Deserialize)]
+struct ResumableQuery {
+    /// Present on a session creation request.
+    upload_type: Option<UploadType>,
+    /// Present on a chunk write, offset query, or cancellation.
+    session: Option<String>,
```
jan-auer: It looks like we're only ever using `ResumableQuery` to classify the response, but we never actually read the session from it. In this case, to avoid an allocation, we could use [`Option<IgnoredAny>`](https://docs.rs/serde/latest/serde/de/struct.IgnoredAny.html).

This is under the assumption that the endpoint handler then extracts `SessionQuery` again. For completeness, we could also pass the session as additional state into the endpoint handler, but I don't think that makes the implementation cleaner.

objectstore-server/src/endpoints/resumable.rs
```diff
@@ -0,0 +1,481 @@
...
+    value
+        .to_str()
+        .ok()
+        .filter(|v| v.bytes().all(|b| b.is_ascii_digit()))
+        .and_then(|v| v.parse().ok())
+        .ok_or_else(|| ApiError::Client(format!("{HEADER_UPLOAD_LENGTH} must be a byte count")))
+}
+
+/// Reads the required [`HEADER_UPLOAD_OFFSET`] header.
+fn upload_offset(headers: &HeaderMap) -> ApiResult<UploadOffset> {
```
jan-auer: As an idea, it could be nice to implement this as an extractor type also. Similar to how [`TypedHeader`](https://docs.rs/axum-extra/latest/axum_extra/struct.TypedHeader.html) works, just that we need a non-standard header here.

objectstore-server/src/endpoints/resumable.rs
```diff
@@ -0,0 +1,481 @@
...
+        .map_err(|e: objectstore_types::resumable::InvalidUploadOffset| {
+            ApiError::Client(e.to_string())
+        })
+}
+
+/// Reads the required `Content-Length` header.
+///
+/// Chunks declare their length so the server can forward only the prefix a backend accepts
+/// without buffering the body to find out how long it is.
+fn content_length(headers: &HeaderMap) -> ApiResult<u64> {
```
jan-auer: For this, we should be able to use `TypedHeader<ContentLength>` directly. Please check if this improves readability if we add this to the endpoint argument list rather than the code.

jan-auer (commented):
Preliminary review.

objectstore-service/src/backend/common.rs
jan-auer: In code, the original intent was to distinguish these cases more clearly (also so that we can log errors for one but not the other). Since it is a valid response for a backend to reject a resumable upload, I think we should return a non-error in this case and return something like `Ok(None)`. This can then be cleanly mapped to a distinct status code in the web layer. This creates a basically self-documenting interface.

objectstore-types/src/resumable.rs
```diff
@@ -44,82 +22,19 @@ const OFFSET_WILDCARD: &str = "*";
...
-    where
-        D: Deserializer<'de>,
-    {
-        let s = String::deserialize(deserializer)?;
-        Self::new(s).map_err(serde::de::Error::custom)
-    }
-}
+/// The token is an opaque identifier whose contents are defined and interpreted by the storage
+/// backend.
+pub type SessionToken = String;
```
jan-auer: As a suggestion, can we make this a newtype wrapper, if needed with `AsRef` and other implementations? This makes it easier to enforce the type across the entire codebase and add further restrictions at any point.

objectstore-types/src/resumable.rs
```diff
@@ -1,35 +1,13 @@
 //! Types for the resumable upload protocol.
```
jan-auer: Let's add a bit more documentation here. This is also pulled into the client, so someone reading client docs may come here and expect more information about resumable uploads. 

If we have duplicate docs in the server, we can refer here, although I think it's fine to have a bit of duplication between the endpoint module and here.

objectstore-server/src/endpoints/objects.rs
```diff
@@ -15,31 +16,79 @@ use serde::Serialize;
...
+}
+
 /// Response returned when inserting an object.
 #[derive(Debug, Serialize)]
 pub struct InsertObjectResponse {
     pub key: String,
 }
 
-async fn objects_post(
+async fn create_object(
```
jan-auer, line 85: _nit_: I'd leave the original method names as they clearly describe what endpoint they are called in.

objectstore-server/tests/resumable.rs
```diff
@@ -0,0 +1,322 @@
...
+/// Sends a raw HTTP/1.1 `PUT`, preserving the caller's exact body framing headers.
+async fn raw_put(server: &TestServer, path: &str, headers: &str, body: &str) -> Result<String> {
+    let url = reqwest::Url::parse(&server.url(path))?;
+    let host = url
+        .host_str()
+        .expect("test server URL has a host")
+        .to_owned();
+    let port = url
+        .port_or_known_default()
+        .expect("test server URL has a port");
```
jan-auer: These we can expose from `TestServer` directly as methods.

objectstore-service/src/backend/tiered.rs
```diff
@@ -96,6 +96,18 @@
 //! already-mutated state and still returns `true` — so callers do not mistakenly
 //! treat a successful commit as a lost race and clean up data that was actually
 //! persisted.
+//!
+//! # Resumable Uploads
```
jan-auer, line 100: Let's remember to update this section when we implement this. Consider adding a TODO?

objectstore-service/src/resumable.rs
```diff
@@ -0,0 +1,36 @@
...
+
+/// How far a resumable upload has progressed.
+///
+/// Returned by both
+/// [`Backend::put_chunk`](crate::backend::common::Backend::put_chunk) and
+/// [`Backend::upload_offset`](crate::backend::common::Backend::upload_offset), because an
+/// offset query commits an object that was assembled but not yet committed and therefore
+/// has the same two outcomes as a chunk write.
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+pub enum UploadProgress {
```
jan-auer: This looks like a useful type for the Rust client, too. Move to objectstore-types?

objectstore-service/src/backend/common.rs
lcian: I'm fine with returning `Ok(None)`, however I cannot find a suitable status code to distinguish the two, I've reviewed all of them and none seems to cleanly fit the semantics we want of "supported but rejected for this request".

objectstore-server/src/endpoints/objects.rs
lcian: The new methods are also called from the same endpoints.

objectstore-server/src/endpoints/resumable.rs
lcian: Implemented it for all headers. This `TypedHeader` is pretty nice. We can actually use directly by just implementing `headers_core::Header` for our own header types.

objectstore-service/src/backend/common.rs
lcian: We're still going to return 501 Not Implemented.

objectstore-server/src/endpoints/objects.rs
```diff
@@ -15,31 +16,80 @@ use serde::Serialize;
...
+    State(state): State<ServiceState>,
+    target: Option<ResumableTarget>,
+    request: Request,
+) -> Response {
+    match target {
+        Some(ResumableTarget::NewSession) => resumable::create_session.call(request, state).await,
+        Some(ResumableTarget::ExistingSession) => {
+            ApiError::Client("`session` requires an object key; use PUT on the object path".into())
+                .into_response()
+        }
```
jan-auer: For `post` and `delete` we could dispatch based on `target.is_some()`. `create_session` internally will then fail if there's no session.

objectstore-server/src/resumable.rs
```diff
@@ -0,0 +1,319 @@
...
+    session: Option<IgnoredAny>,
+}
+
+impl ResumableQuery {
+    /// Classifies a request that may create a session or act on one.
+    fn classify(self) -> Option<ResumableTarget> {
+        match (self.upload_type, self.session) {
+            // A session unambiguously selects an existing upload, so a redundant upload_type can
+            // be ignored rather than making the request ambiguous.
+            (Some(_), Some(_)) | (None, Some(_)) => Some(ResumableTarget::ExistingSession),
```
jan-auer:
```suggestion
            (_, Some(_)) => Some(ResumableTarget::ExistingSession),
```

objectstore-test/src/server.rs
```diff
@@ -98,12 +98,22 @@ impl TestServer {
         Self::with_config(Config::default()).await
     }
 
+    /// Returns the IP address on which the server is listening.
+    pub fn host(&self) -> IpAddr {
+        self.socket.ip()
```
jan-auer, line 103: Note that this is now different than what `self.url()` returns.

jan-auer (commented):
✅  LGTM!

lcian (approved):

PR #595, opened 2026-08-07
feat(cogs): report backend changes to a change stream

matt-codecov:
Ref FS-210

In GCS and Bigtable `Backend` impls, use the `ChangeStream` impl to record write/update/delete events.

Other backends are left as a TODO for the moment: Ref FS-489

Currently Objectstore lacks the config plumbing needed to instantiate anything other than `NoopStream`. That will come in later PRs.

### DIFF
--- a/objectstore-service/src/backend/bigtable.rs
+++ b/objectstore-service/src/backend/bigtable.rs
@@ -469,11 +469,15 @@ fn bumped_tti_metadata(metadata: &Metadata) -> Metadata {
 }
 
 /// Builds the three mutations that write an object row: clear existing data,
-/// then set the payload and metadata cells.
+/// then set the payload and metadata cells. Returns them with the resulting row size.
 ///
 /// Used by both [`BigTableBackend::put_row`] (unconditional write) and
 /// [`BigTableBackend::put_non_tombstone`] (conditional write).
-fn object_mutations(mut metadata: Metadata, payload: Vec<u8>) -> Result<[v2::Mutation; 3]> {
+fn object_mutations(
+    path: &[u8],
+    mut metadata: Metadata,
+    payload: Vec<u8>,
+) -> Result<([v2::Mutation; 3], u64)> {
     let (family, timestamp_micros) = match metadata.time_expires {
         None => (FAMILY_MANUAL, -1),
         Some(deadline) => (FAMILY_GC, system_time_to_micros(deadline)?),
@@ -485,7 +489,7 @@ fn object_mutations(mut metadata: Metadata, payload: Vec<u8>) -> Result<[v2::Mut
     let metadata_bytes = serde_json::to_vec(&metadata)
         .map_err(|cause| Error::serde("failed to serialize metadata", cause))?;
 
-    Ok([
+    let mutations = [
         // NB: We explicitly delete the row to clear metadata on overwrite.
         delete_row_mutation(),
         mutation(mutation::Mutation::SetCell(mutation::SetCell {
@@ -500,7 +504,33 @@ fn object_mutations(mut metadata: Metadata, payload: Vec<u8>) -> Result<[v2::Mut
             timestamp_micros,
             value: metadata_bytes,
         })),
-    ])
+    ];
+
+    let size = row_size(path, &mutations);
+    Ok((mutations, size))
+}
+
+/// Approximates the bytes a row occupies, as its key plus every cell value written.
+///
+/// This function does not distinguish between object rows and tombstone rows. It does not
+/// include Bigtable's own overhead.
+fn row_size(path: &[u8], mutations: &[v2::Mutation]) -> u64 {
+    let cells: usize = mutations
+        .iter()
+        .filter_map(|m| match &m.mutation {
+            Some(mutation::Mutation::SetCell(cell)) => Some(cell.value.len()),
+            _ => None,
+        })
+        .sum();
+
+    (path.len() + cells) as u64
+}
+
+/// The moment a row written now under `policy` is expected to be reclaimed.
+///
+/// Returns `None` for [`ExpirationPolicy::Manual`], which never expires on its own.
+fn expiry_from_policy(policy: ExpirationPolicy, now: SystemTime) -> Option<SystemTime> {
+    policy.expires_in().map(|ttl| now + ttl)
 }
 
 /// Metadata carried by tombstone rows in the `t` (tombstone-meta) column.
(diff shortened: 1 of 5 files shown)


objectstore-service/src/backend/gcs.rs
```diff
@@ -871,7 +897,12 @@ impl Backend for GcsBackend {
 
             Ok(())
         })
-        .await
+        .await?;
```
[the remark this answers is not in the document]
matt-codecov: this is harmless outside of the incredibly unlikely case of a racing delete/write being sequenced differently in Objectstore and Kafka. but i will fix

objectstore-service/src/backend/bigtable.rs
```diff
@@ -505,6 +505,29 @@ fn object_mutations(mut metadata: Metadata, payload: Vec<u8>) -> Result<[v2::Mut
     ])
 }
 
+/// Approximates the bytes a row occupies, as its key plus every cell value written.
+///
+/// This function does not distinguish between object rows and tombstone rows. It does not
+/// include Bigtable's own overhead.
+fn row_bytes(path: &[u8], mutations: &[v2::Mutation]) -> u64 {
```
jan-auer: _nit_: Let's use `length` or `size` since `bytes` is commonly used to refer to the actual data.

objectstore-service/src/backend/bigtable.rs
```diff
@@ -943,8 +1003,10 @@ impl Backend for BigTableBackend {
             payload.push(chunk);
         }
 
-        self.put_row(path, metadata.clone(), payload.into_bytes().into(), "put")
-            .await?;
+        // Inline `put_row()` because we need the mutations to compute their size.
+        let mutations = object_mutations(metadata.clone(), payload.into_bytes().into())?;
+        self.mutate(path.clone(), mutations.clone(), "put").await?;
+        self.report_write(id, &path, &mutations, metadata.time_expires);
```
jan-auer: You can make `object_mutations` instead return the size alongside the mutations, return that from put_row, and then use the returned value to report the write. This yields all data in the correct sequence, removes the inlining, and mutations no longer have to be cloned.

objectstore-service/src/backend/gcs.rs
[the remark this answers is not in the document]
jan-auer: Easy to fix by returning a boolean from the retry closure.

objectstore-service/src/backend/gcs.rs
```diff
@@ -728,6 +745,11 @@ impl Backend for GcsBackend {
         objectstore_log::debug!("Writing to GCS backend");
         let gcs_metadata = GcsObject::from_metadata(metadata);
 
+        // The payload arrives as a stream with no declared length, so the stored size is
+        // only known once the upload has drained. Safe from double-counting because this
+        // request is not retried.
+        let (stored_size, stream) = counting_stream(stream);
```
jan-auer: Please double-check, but GCS should be returning a [`X-Goog-Stored-Content-Length`](https://docs.cloud.google.com/storage/docs/json_api/v1/parameters#xgoogstoredcontentlength) from the upload request which we can use. Even better, this is authoritative. 

Just know that this doesn't include metadata, and GCS charges for metadata as bytes, too.

jan-auer (commented):
Preliminary review

objectstore-service/src/backend/gcs.rs
matt-codecov: TODO also count metadata

objectstore-service/src/backend/bigtable.rs
```diff
@@ -1117,6 +1175,7 @@ impl HighVolumeBackend for BigTableBackend {
                 .await?;
 
             if write_succeeded {
+                self.change_stream.delete(id);
```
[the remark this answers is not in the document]
jan-auer: delete of non-existent object is idempotent for both COGS and GC, although it would be semantically cleaner if we didn't emit in this case.

objectstore-service/src/backend/gcs.rs
```diff
@@ -757,16 +800,45 @@ impl Backend for GcsBackend {
...
+        match stored_size {
+            Some(payload_size) => {
+                self.change_stream.write(
+                    id,
+                    payload_size + gcs_metadata.metadata_size(),
+                    metadata.time_expires,
+                );
+            }
+            None => {
+                objectstore_metrics::count!("cost_tracker.unreported", reason = "no_stored_size");
```
jan-auer: This is a general log for the change stream and all of its subscribers, not just the cost tracker.

objectstore-service/src/backend/gcs.rs
```diff
@@ -757,16 +800,45 @@ impl Backend for GcsBackend {
...
+                response.drain_body().await;
+                Some(size)
+            }
+            None => response
+                .json::<GcsObject>()
+                .await
+                .ok()
+                .and_then(|object| object.size)
+                .and_then(|size| size.parse().ok()),
+        };
```
jan-auer: Is there an indication in that we'd get one but not the other? If not, we can simplify here and keep just one of the branches.

If you'd like to keep this, as a suggestion, let's move all of this into the `stored_content_length` helper, so the function here reads better.

objectstore-service/src/backend/gcs.rs
```diff
@@ -757,16 +800,45 @@ impl Backend for GcsBackend {
...
+
+        match stored_size {
+            Some(payload_size) => {
```
jan-auer: _nit_: This can be shortened with `if let Some()`

objectstore-service/src/backend/gcs.rs
```diff
@@ -842,36 +914,43 @@ impl Backend for GcsBackend {
...
+            .await?;
+
+        if deleted {
+            self.change_stream.delete(id);
+        }
+
+        Ok(())
     }
 
     async fn join(&self) {
```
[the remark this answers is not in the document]
jan-auer: This is OK, we'll replace multipart soon.

jan-auer (approved):
LGTM

objectstore-service/src/backend/gcs.rs
matt-codecov: the emulator didn't send the header. but yeah i can move into the helper

objectstore-service/src/backend/bigtable.rs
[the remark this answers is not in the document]
matt-codecov: because this is harmless, i cut a ticket (FS-492) and referred to it in a comment so i can merge the stack and then follow up. i don't want to touch core logic recklessly

PR #597, opened 2026-08-11
feat(cogs): publish backend change streams to Kafka

matt-codecov:
Ref FS-210

Add `CostTrackerConfig` struct to `objectstore-server`'s config. This allows storage COGS to be configured end-to-end; `CostTrackerConfig` on `objectstore-server`'s config and `CostTrackerStreamConfig` on each `Backend` impl will produce a `CostTrackerStream` that holds a shared Kafka producer and streams change messages on the `shared-resources-inventory` topic.

While `ChangeStream` is abstract, the abstraction is not exposed in config. It's assumed each `ChangeStream` impl will name itself explicitly in dedicated config sections.

---

<sub>Stack created with <a href="https://github.com/github/gh-stack">GitHub Stacks CLI</a> • <a href="https://gh.io/stacks-feedback">Give Feedback 💬</a></sub>

### DIFF
--- a/objectstore-server/src/config.rs
+++ b/objectstore-server/src/config.rs
@@ -41,6 +41,7 @@ use std::time::Duration;
 use anyhow::Result;
 use figment::providers::{Env, Format, Serialized, Yaml};
 use objectstore_service::backend::local_fs::FileSystemConfig;
+use objectstore_service::change_stream::CostTrackerConfig;
 use objectstore_types::auth::Permission;
 use secrecy::{CloneableSecret, SecretBox, SerializableSecret, zeroize::Zeroize};
 use serde::{Deserialize, Serialize};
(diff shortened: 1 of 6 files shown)


objectstore-server/src/state.rs
```diff
@@ -62,8 +62,13 @@ impl Services {
...
         tokio::spawn(track_allocator_metrics(config.runtime.metrics_interval));
 
-        let backend =
-            backend::from_config(config.storage.clone(), &ChangeStreamFactory::default()).await?;
+        // Absent config means no transport, and therefore no reporting.
+        let streams = config
+            .storage_cogs
+            .as_ref()
+            .map(ChangeStreamFactory::new)
+            .unwrap_or_default();
```
jan-auer, line 70: _nit_: Reading it like this, implementing `Default for ChangeStreamFactory` leaves the question of what the default actually does. There's no natural "default" for change streams. Maybe we should remove that and explicitly instantiate the noop factory here.

jan-auer (approved):

objectstore-server/src/config.rs
```diff
@@ -868,6 +895,67 @@ mod tests {
...
+            let StorageConfig::BigTable(storage) = &dbg!(&config).storage else {
+                panic!("expected bigtable storage");
+            };
+            let stream = storage.change_stream.as_ref().expect("change stream");
+            assert_eq!(stream.shared_resource_id, "bigtable_objectstore");
+            assert_eq!(stream.listeners.kafka.as_ref().unwrap().sample_rate, 0.5);
+
+            let sink = config.change_stream.kafka.as_ref().expect("kafka sink");
+            assert_eq!(sink.topic, "my-topic");
+            assert_eq!(sink.bootstrap_servers, ["kafka:9092"]);
```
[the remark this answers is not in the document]
jan-auer: This is true, but I think we can accept this.

objectstore-server/src/state.rs
matt-codecov: there isn't a separate noop factory, there's just `ChangeStreamFactory`. the way i think of it is: the default factory has no sink configs and only knows how to build `NoopStream`. if you have sink configs, you build a non-default factory with them and that factory knows how to build other `ChangeStream`s

PR #599, opened 2026-08-11
feat(config): resolve ${file:...} and ${VAR} references in config

matt-codecov:
Add `serde-vars` crate to allow `${file:...}` and `${VAR}` to be used as values in our config for deferred loads.

```yaml
storage_cogs:
  type: kafka
  override_params:
    sasl.password: ${file:/var/secrets/kafka-password}
    # sasl.password: ${KAFKA_PASSWORD} # alternate
```

This is helpful for a few reasons:
- YAML config can be complete. Static YAML can name secrets/runtime values while deferring the actual loading.
- We don't need to write bespoke file I/O for secrets/configs that are mounted as files (e.g. GCP secrets).
- Edge case: the `sasl.password` config key above.

The `override_params` map is passed verbatim to `rdkafka` and `rdkafka` expects keys such as `sasl.password`. That value is obviously secret, so we can't include it in our config YAML. However, because the config key contains a `.`, we can't set it with an environment variable and can _only_ set it in YAML. `serde-vars` lets us configure the secret in YAML while deferring the actual loading of the secret value until later.

---

<sub>Stack created with <a href="https://github.com/github/gh-stack">GitHub Stacks CLI</a> • <a href="https://gh.io/stacks-feedback">Give Feedback 💬</a></sub>

### DIFF
--- a/objectstore-server/src/config.rs
+++ b/objectstore-server/src/config.rs
@@ -30,6 +30,29 @@
 //!   type: filesystem
 //!   path: /data
 //! ```
+//!
+//! # Variable References
+//!
+//! Any configuration value may be written as a reference, which is resolved after all
+//! sources have been merged. `${file:PATH}` is replaced by that file's contents and
+//! `${VAR_NAME}` by that environment variable:
+//!
+//! ```yaml
+//! storage_cogs:
+//!   type: kafka
+//!   override_params:
+//!     sasl.password: ${file:/var/secrets/kafka-password}
+//! ```
+//!
+//! A reference must be the entire value: `${A}` works, `prefix-${A}` does not. Referencing
+//! a file that cannot be read, or an environment variable that is not set, is an error at
+//! startup.
+//!
+//! ## Relationship to `OS__` environment variables
+//!
+//! These are different mechanisms and neither replaces the other. They can even be used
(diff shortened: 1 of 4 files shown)


objectstore-server/src/config.rs
```diff
@@ -30,6 +30,39 @@
 //!   type: filesystem
 //!   path: /data
 //! ```
+//!
+//! # Variable References
```
jan-auer, line 34: This is rather verbose and seems written too specific for secrets. Let's describe the general mechanism briefly but cut out what's redundant.

objectstore-server/src/config.rs
```diff
@@ -956,6 +1013,172 @@ mod tests {
...
+                config.sentry.dsn.unwrap().expose_secret().as_str(),
+                "https://public@example.invalid/1"
+            );
+
+            Ok(())
+        });
+    }
+
+    #[test]
+    fn a_value_that_is_not_a_reference_is_left_alone() {
```
jan-auer: These tests are rather verbose and repetitive. Let's try to merge what we can (e.g. assert an existing and a missing value in the same test).

jan-auer (approved):

objectstore-server/src/config.rs
matt-codecov: i suppose the PR description is where the sales pitch for `serde-vars` belongs and the in-code docs should just describe the mechanism as you suggest. will change

PR #600, opened 2026-08-26
chore(gocd): use GitHub App credentials

joshuarli:
Replace GoCD GitHub token references with GitHub App credentials.

- Pass GITHUB_APP_ID and GITHUB_APP_PRIVATE_KEY to GoCD jobs.
- Preserve legacy runtime compatibility where the existing helper supports it.


jan-auer (approved):
Unblocking for when this is ready to merge

kenzoengineer (approved):

PR #606, opened 2026-08-31
ref(service): Simplify semantic service errors

jan-auer:
Replace the service error enum with an opaque error and semantic error kinds.

Preserve backend response details as structured source errors. Classify panics, client stream failures, corrupt data, and capacity failures consistently across service and server code. The server now only needs to match on the public "kind", but not on the full enum with potentially nested errors.

These errors still don't capture backtraces, so we rely on the error kind + manual context to locate them precisely.

Replaces #547
Closes FS-446
Ref FS-358


jan-auer:
After iterating on this a bit more, this approach doesn't work at all. Since backtraces are not stabilized, I'm not able to find a good API that allows us to capture them with the tracing integration.

We will instead go back to tracking custom context descriptions for now.

objectstore-service/src/backend/gcs.rs
```diff
@@ -522,7 +499,9 @@ impl GcsBackend {
 
         Ok(Self {
             client: common::reqwest_client(),
-            endpoint: endpoint_str.parse().context("invalid GCS endpoint URL")?,
+            endpoint: endpoint_str
+                .parse()
+                .map_err(|error| anyhow::anyhow!("invalid GCS endpoint URL: {error}"))?,
```
jan-auer: Check if we can bring back anyhow's context for this, rather than formatting the inner error like this.

objectstore-service/src/error.rs
```diff
@@ -1,177 +1,68 @@
...
-    /// [`HighVolumeBackend`](crate::backend::common::HighVolumeBackend) methods.
-    #[error("unexpected tombstone")]
-    UnexpectedTombstone,
-
-    /// The requested byte range is not satisfiable for the object's size.
-    #[error("range not satisfiable (object size: {total} bytes)")]
+/// The client-visible semantic classification of a service error.
+#[non_exhaustive]
+#[derive(Clone, Copy, Debug, Eq, PartialEq)]
+pub enum ErrorKind {
```
jan-auer, line 43: Match original order of variants.

objectstore-service/src/error.rs
```diff
@@ -180,76 +71,302 @@ pub enum Error {
...
-            Self::Reqwest { .. } => Level::ERROR,
-            Self::BackendResponse { .. } => Level::ERROR,
-            Self::GcpAuth(_) => Level::ERROR,
-            Self::Panic(_) => Level::ERROR,
-            Self::Dropped => Level::ERROR,
-            Self::UnexpectedTombstone => Level::ERROR,
-            Self::InvalidUploadId(_) => Level::DEBUG,
-            Self::Generic { .. } => Level::ERROR,
+        match self.kind {
+            ErrorKind::InvalidMetadata
```
jan-auer: Bring back separate match arms like before - also port over comments that are still valid.

objectstore-service/src/backend/extensions.rs
```diff
@@ -41,6 +57,28 @@ impl SendTraced for reqwest::RequestBuilder {
     }
 }
 
+/// Classifies a reqwest transport result while leaving successful values available for
+/// caller-specific handling.
+pub(super) fn classify_transport_result<T>(
+    result: reqwest::Result<T>,
+    context: &'static str,
+) -> Result<T> {
+    result.map_err(|source| {
```
jan-auer: Instead of calling `map_err` here, make this a function we can call from _inside_ `map_err`. That way, we can preserve the current chaining.

Also, check if we can remove `From<reqwest::Error> for Error`, because that ensures on type level that we have to call this.

objectstore-server/examples/capture_service_error.rs
```diff
@@ -0,0 +1,88 @@
+//! Sends a production-shaped service error to Sentry and waits for delivery.
```
jan-auer: Remove this.

objectstore-service/src/service.rs
```diff
@@ -260,7 +260,7 @@ impl StorageService {
         id: ObjectId,
         metadata: Metadata,
     ) -> Result<InitiateMultipartResponse> {
-        metadata.validate()?;
+        metadata.validate().kind(ErrorKind::InvalidMetadata)?;
```
lcian, line 263: This should have kind `Internal` or something that equivalently maps to a 500 Internal Server Error, as it's validating an internal invariant on `Metadata` , so it's not fault of the client if this is violated.

objectstore-server/src/endpoints/common.rs
```diff
@@ -90,23 +130,28 @@ impl ApiError {
...
+            ApiError::Service(error) => match error.kind() {
+                ServiceErrorKind::InvalidMetadata
+                | ServiceErrorKind::InvalidUploadId
+                | ServiceErrorKind::ClientStream
+                | ServiceErrorKind::UnknownUploadSession
+                | ServiceErrorKind::ChunkExceedsUploadLength { .. } => StatusCode::BAD_REQUEST,
+                ServiceErrorKind::RangeNotSatisfiable { .. } => StatusCode::RANGE_NOT_SATISFIABLE,
+                ServiceErrorKind::UploadOffsetMismatch { .. } => StatusCode::CONFLICT,
+                ServiceErrorKind::UploadSessionGone => StatusCode::GONE,
+                ServiceErrorKind::AtCapacity => StatusCode::TOO_MANY_REQUESTS,
```
lcian, line 142: Not strictly related to this PR but I would argue this should map to StatusCode::SERVICE_UNAVAILABLE instead, as it's not a rate limit per se.

jan-auer: I would like to review the status codes for rate limits more closely in a follow-up. It can make sense that we return 503 here, but we should double-check if we do actually want to return 503 for `BackendTimeout` and `BackendUnavailable`.

lcian: Okay. fwiw I think we do.

lcian (approved):
Looks good!

PR #607, opened 2026-08-31
feat(service): Instrument the remaining backends with the change stream

matt-codecov:
- document the `ChangeStream` abstraction in `objectstore-service/docs/architecture.md`
- wire up `LocalFsBackend`, `InMemoryBackend`, and `S3CompatibleBackend` with a `ChangeStreamFactory` / `self.change_stream` handle
- make the above backends emit write/update/delete messages to `self.change_stream`

Closes FS-489

also includes a drive-by fix: previously the S3 backend would return `Err(...)` when deleting a non-existent object. the rest of our backends return `Ok(())` in that case. so this PR makes the S3 backend match the other backends' behavior


objectstore-service/docs/architecture.md
```diff
@@ -179,6 +173,48 @@ sampling rate. When aggregating, divide each row's value by its `sample_rate`.
...
+[storage COGS change stream](#storage-cogs) is configured with a service-wide
+[`CostTrackerConfig`](change_stream::CostTrackerConfig) and per-backend
+[`CostTrackerStreamConfig`](change_stream::CostTrackerStreamConfig)s. These
+configurations are connected in [`ChangeStreamFactory`](change_stream::ChangeStreamFactory)
+to build a [`CostTrackerStream`](change_stream::CostTrackerStream).
+
+New `ChangeStream` implementations may follow the same pattern:
+- per-backend configuration for per-backend IDs or configuration
+- service-wide configuration for a stream sink
+- glue code in and around `ChangeStreamFactory`
```
aldy505, line 216: Not sure what you meant by this

objectstore-service/src/backend/local_fs.rs
```diff
@@ -51,18 +55,37 @@ pub struct FileSystemConfig {
...
     path: PathBuf,
+
+    change_stream: Arc<dyn ChangeStream>,
 }
 
 impl LocalFsBackend {
     /// Creates a new [`LocalFsBackend`] rooted at the directory in `config`.
-    pub fn new(config: FileSystemConfig) -> Self {
-        Self { path: config.path }
+    pub fn new(config: FileSystemConfig, streams: &ChangeStreamFactory) -> Self {
```
aldy505, line 83: Why are we using `&ChangeStreamFactory` here instead of `Arc<dyn ChangeStream>`? Shouldn't `ChangeStream` implementation is dynamically switchable, in SaaS we can use the cost tracker, in self-hosted we can use the garbage collector?

objectstore-service/src/backend/s3_compatible.rs
```diff
@@ -52,6 +57,19 @@ pub struct S3CompatibleConfig {
...
+    /// # Default
+    ///
+    /// `None`, which disables reporting for this backend.
+    ///
+    /// # Environment Variables
+    ///
+    /// - `OS__STORAGE__COGS__SHARED_RESOURCE_ID=s3_objectstore`
+    /// - `OS__STORAGE__COGS__SAMPLE_RATE=1.0` (optional)
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub cogs: Option<CostTrackerStreamConfig>,
```
lcian, line 72: I understand the design with the `ChangeStreamFactory` now.
That allows us to define a `pub cleanup: Option<CleanupStreamConfig>` for expiry cleanup just here but not on e.g. the GCS backend, making it impossible to construct a GCS backend with a `CleanupChangeStream`.

If we were passing in `Arc<dyn ChangeStream>` to each backend instead, we wouldn't be able to enforce that as easily.

objectstore-service/src/backend/local_fs.rs
lcian: I was also thinking about this. See https://github.com/getsentry/objectstore/pull/607/changes#r3952214185 for my though process. Matt will probably have more to add to this as well.

objectstore-service/docs/architecture.md
lcian: If you want to add a new `ChangeStream` implementor to certain backends, you need to change:
- The Service config to configure where the changes get written for that particular change stream (for COGS this is the `storage_cogs` top-level config key in the yaml -- you do this at the Service level so you can share a single Kafka producer/db connection/etc.)
- The backend config for the backends that should support such change stream implementation (in this PR we needed to add the `cogs` field to `FileSystemConfig`)
- `ChangeStreamFactory::new` and `::build`, to teach the factory how to use those new configs.

objectstore-service/docs/architecture.md
```diff
@@ -179,6 +173,48 @@ sampling rate. When aggregating, divide each row's value by its `sample_rate`.
...
+  unchanged. In practice this is a TTI bump.
+- `delete(id)`: `id` was deleted explicitly.
+
+The stream describes physical storage per backend. When using
+[`TieredStorage`](backend::tiered::TieredStorage), objects that are stored in
+long-term storage will emit a change record for the actual object in long-term
+storage as well as for the tombstone record in high-volume storage.
+
+`size` is a count of bytes that the backend actually stores for an object. This
+includes object payloads, metadata, and sometimes backend-specific overhead.
```
lcian, line 194: Nit: should we say that `size` should be an _estimate_ or _proportional_ to the bytes that you're storing/billed for in/by the backend, rather than literally the size?
Why: looking at S3 docs it seems that the storage bytes you're charged for has some nuances that depend on things such as the storage tier: https://aws.amazon.com/s3/pricing/#:~:text=*%20S3%20Intelligent%2DTiering,pricing%20page.

lcian (approved):
Looks good to me.

objectstore-service/docs/architecture.md
aldy505: Oh i was specifically asking for the last point, why (and when) do we need ChangeStreamFactory?

lcian: The reason for `ChangeStreamFactory` is to construct `Arc<dyn ChangeStream>` in a more flexible way, for example right now we pass in the Kafka producer in `ChangeStreamFactory::new` and share it every time `build` is called with the COGS config.
If we want to support a new implementor of `ChangeStream`, the `ChangeStreamFactory::new` and `::build` need to be adapted to take in the configs for that new implementor and output a `ChangeStream` that potentially combines `cogs` and this new implementor as the resulting `Arc<dyn ChangeStream>`.

matt-codecov: looking at it another way: we don't want to copy-paste the Kafka connection details onto every backend's config, and we also don't want each backend to instantiate their own heavy Kafka producers. so we configure the Kafka connection and create the producer at a service-wide level. however, backends may have backend-specific change stream configs (for COGS it's the `shared_resource_id` and the sample rate), and we want those details to live inside the config for the backend rather than a separate location that we have to keep synced. so, the two halves of the change stream config for COGS are in very different places, the `ChangeStreamFactory` is the thing that knows how to join them together.

objectstore-service/src/backend/s3_compatible.rs
matt-codecov: if somebody wanted to configure a `CleanupChangeStream` in GCS i'd be fine with that, actually. maybe they've got a pre-existing or shared bucket with its own use of `Custom-Time` or lifecycle policies and they need to implement GC separately

originally each backend just had a list of `ChangeStreamConfig`s, and `ChangeStreamConfig` had an `Option<CostTrackerStreamConfig>` and we'd have added an `Option<CleanupChangeStreamConfig>` with the expectation that each `ChangeStreamConfig` have exactly one non-null member populated. basically an enum but serialization-friendly. i flattened to this to address freedback from Jan. i've already forgotten the rationale lol

objectstore-service/docs/architecture.md
matt-codecov: imo it should literally be the size, and nuances like graduated pricing should be handled by a downstream COGS-specific consumer

objectstore-service/src/backend/s3_compatible.rs
```diff
@@ -416,9 +472,14 @@ impl<T: TokenProvider> Backend for S3CompatibleBackend<T> {
             .await?
             .drain_body()
             .await;
+        self.change_stream.delete(id);
```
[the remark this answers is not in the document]
lcian: Seems legit, please check.

objectstore-service/docs/architecture.md
```diff
@@ -146,15 +146,16 @@ rate-limiting failures at a higher layer) are not counted.
 
 This is gated behind the `storage_cogs` Cargo feature.
```
lcian, line 147:
```suggestion
This is gated behind the `storage-cogs` Cargo feature.
```

objectstore-service/docs/architecture.md
```diff
@@ -146,15 +146,16 @@ rate-limiting failures at a higher layer) are not counted.
...
-[`CostTrackerStreamConfig`](change_stream::CostTrackerStreamConfig) and the
-service has a usable transport for it, and unless the `storage-cogs` feature is
-compiled in at all.
+Each backend reports every write/overwrite, applied expiry extension, and delete
+it performs on stored objects to a [`ChangeStream`](change_stream::ChangeStream)
+(see [the change stream section](#change-streams)). To turn this change stream
+into COGS data, a stream consumer has to merge each change event into an
+external table to update an inventory of objects. The inventory table can be
+queried to break down each backend's storage utilization by `app_feature`. To
+enable storage COGS, enable the `storage_cogs` Cargo feature and provide a
```
lcian, line 155:
```suggestion
enable storage COGS, enable the `storage-cogs` Cargo feature and provide a
```

objectstore-service/src/backend/in_memory.rs
```diff
@@ -42,6 +43,18 @@ impl StoreEntry {
...
     }
+
+    /// Number of bytes an entry occupies (payload plus serialized metadata)
+    pub(crate) fn stored_size(&self) -> usize {
+        match self {
+            StoreEntry::Object(metadata, payload) => json_len(metadata) + payload.len(),
+            // A tombstone carries no payload, only its redirect target. Its expiry is
+            // held out of band, so it contributes no bytes.
+            StoreEntry::Tombstone(tombstone) => {
+                tombstone.target.as_storage_path().to_string().len()
```
lcian, line 54: Should expiry really not count? We store it on the tombstone.

lcian (approved):
See bot and my comments, otherwise looks good.

objectstore-service/src/backend/s3_compatible.rs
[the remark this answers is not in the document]
matt-codecov: AWS returns 204 whether the object existed or not. if we want to tell whether the object deletion actually occurred, we have to rely on behavior that minio and probably other backends don't implement (`If-Match` predicates forcing a 412). i'll leave a comment about it; the cost of a spurious kafka message is very low compared to the cost of guessing how backends will handle `If-Match` and hoping they never interfere with the actual execution of the delete request

PR #609, opened 2026-09-01
feat(gcs): Add resumable uploads

lcian:
Implements the Resumable Upload API on the GCS backend by forwarding to the upstream API.

2 notable things:
- Resumable session tokens will now be encrypted with AES-256-GCM when Objecstore is configured with compatible symmetric keys. This is necessary for GCS, where passing through the session token verbatim would allow direct operations on the bucket.
- In the previous PR/design, we didn't consider what happens when the user tries to upload a chunk at an offset lower than the persisted one.
GCS supports that, and just ignores the already persisted prefix. Objectstore could have a stricter contract where we enforce the client-provided offset to match what the backend has, but that would require a preflight request to query the offset on every chunk upload.
Instead, we document that Objecstore may or may not accept such requests, and in the case of GCS we just accept them, which practically means that we always pass through the offset and body without any pre-check.

Ref FS-494


lcian:
bugbot run
@sentry review

lcian:
bugbot run
@sentry review

objectstore-service/src/backend/gcs.rs
```diff
@@ -770,6 +910,187 @@ impl Backend for GcsBackend {
...
+                    response.drain_body().await;
+                    Ok(session_uri)
+                }
+            })
+            .await?;
+
+        ResumableSession::new(id, session_uri, total_length)
+            .into_token()
+            .map(Some)
+    }
```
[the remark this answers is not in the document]
lcian: Okay for now.

lcian:
bugbot run
@sentry review

objectstore-service/src/backend/gcs.rs
```diff
@@ -529,6 +593,62 @@ impl GcsBackend {
...
+                    )
+                    .send_traced()
+                    .await
+                    .map_err(|error| Error::reqwest("GCS: query resumable upload", error))?;
+                self.resumable_progress(id, response, session.total_length)
+                    .await
+            }
+        })
+        .await
+    }
```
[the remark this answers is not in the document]
lcian: `InventoryTracker` derives the same stable `record_id` from the storage key, and the downstream state query groups by `shared_resource_id` and `record_id` and selects the latest values. Repeated WRITE records therefore refresh one inventory object and COGS counts its stored size once.
To be checked with Matt if this is okay, as I'm not sure how we want to handle this at the `InventoryTracker` level of abstraction (without knowing that this is used by COGS aggregation and that job deduplicates).

objectstore-service/src/backend/gcs.rs
```diff
@@ -770,6 +1009,188 @@ impl Backend for GcsBackend {
...
+            .resumable_progress(id, response, session.total_length)
+            .await
+        {
+            Err(error) if is_resumable_offset_gap(&error) => {
+                match self.query_resumable_progress(id, &session).await {
+                    Ok(UploadProgress::Incomplete { offset }) => {
+                        Err(Error::UploadOffsetMismatch { offset })
+                    }
+                    Ok(UploadProgress::Committed) | Err(_) => Err(error),
+                }
```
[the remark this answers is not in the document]
lcian: Fixed in 9289e52. Offset-gap reconciliation now returns Complete when the authoritative status query observes a terminal upload, while preserving UploadOffsetMismatch for incomplete uploads and the original backend error when the status query fails. A focused unit test covers all three outcomes.

lcian:
bugbot run
@sentry review

lcian:
bugbot run
@sentry review

objectstore-service/src/backend/gcs.rs
```diff
@@ -740,6 +846,140 @@ impl fmt::Debug for GcsBackend {
...
+        Error::new(
+            ErrorKind::BackendFailure,
+            "GCS: resumable Range header overflows",
+        )
+    })?;
+    if offset > total_length {
+        return Err(Error::new(
+            ErrorKind::BackendFailure,
+            "GCS: incomplete resumable Range reaches declared upload length",
+        ));
```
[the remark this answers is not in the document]
lcian: We assume that GCS behaves correctly and never returns such an offset in the code paths that call this function, namely those that parse a 308 response.

objectstore-service/src/backend/gcs.rs
```diff
@@ -919,6 +1068,214 @@ impl Backend for GcsBackend {
...
+                    let stored_size = object.size.as_deref().and_then(|size| size.parse().ok());
+                    self.report_object_write(
+                        id,
+                        stored_size,
+                        object.metadata_size(),
+                        object.custom_time,
+                    );
+                }
+                progress
+            })
```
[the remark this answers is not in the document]
lcian: In that case we return `BackendFailure` and the caller is who's responsible to query the offset again.

objectstore-service/src/backend/gcs.rs
```diff
@@ -919,6 +1068,214 @@ impl Backend for GcsBackend {
...
+            .request(Method::PUT, session.session_uri.as_str())
+            .await?
+            .header(header::CONTENT_LENGTH, content_length)
+            .header(header::CONTENT_RANGE, content_range)
+            .body(Body::wrap_stream(stream))
+            .send_traced()
+            .await
+            .reqwest_context("GCS: upload resumable chunk")?;
+
+        range_response_to_upload_progress(&session, response)
```
lcian, line 1186: Here we could be nice to the user and, for example, if they submit a chunk at the wrong offset, query the offset again for them and return a 200 with the correct offset instead of an error.
Turns out that's not very easy to do, also because of the different ways that GCS can report an error that we would have to handle, so I'm not doing that for now.
Right now we recognize NOT_FOUND/GONE/CLIENT_CLOSED_REQUEST, and in other cases we'll return 500 (mapped from `ErrorKind::BackendFailure`).
In such cases the user is responsible to query the offset again.

objectstore-service/src/backend/gcs.rs
```diff
@@ -919,6 +1068,206 @@ impl Backend for GcsBackend {
...
+            .await
+            // If we get a completed object back it means that a previous `put_chunk` request
+            // already observed it and wrote to the `change_stream`. If we were to write to it
+            // again we would double count.
+            .map(|(progress, _)| progress)
```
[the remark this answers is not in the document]
lcian: meant to put this on the offset query not on put_chunk

objectstore-service/src/backend/gcs.rs
```diff
@@ -919,6 +1068,206 @@ impl Backend for GcsBackend {
...
+    }
```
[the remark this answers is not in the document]
lcian: meant to put this on the offset query not on put_chunk

objectstore-service/src/service.rs
```diff
@@ -372,33 +387,55 @@ impl StorageService {
...
     /// # Run-to-completion
     ///
     /// Once called, the operation runs to completion even if the returned future is dropped.
-    /// This matters most for the final chunk, which commits the object.
+    /// This matters most for the final chunk, which completes the upload.
     pub async fn put_chunk(
         &self,
         id: ObjectId,
-        session: SessionToken,
+        session: EncryptedSessionToken,
```
jan-auer, line 433: Let's rename these parameters to `token`.

objectstore-service/src/resumable.rs
```diff
@@ -0,0 +1,356 @@
...
+pub(crate) struct SessionToken {
+    #[serde(
+        serialize_with = "serialize_object_id",
+        deserialize_with = "deserialize_object_id"
+    )]
+    pub(crate) object_id: ObjectId,
+    pub(crate) backend_token: BackendToken,
+}
+
+fn serialize_object_id<S>(id: &ObjectId, serializer: S) -> std::result::Result<S::Ok, S::Error>
```
jan-auer, line 36:
```suggestion
fn serialize_object_id<S>(id: &ObjectId, serializer: S) -> Result<S::Ok, S::Error>
```

objectstore-service/src/resumable.rs
```diff
@@ -0,0 +1,356 @@
...
+pub type BackendToken = String;
+
+/// Structured token encrypted at the service boundary.
+#[derive(Deserialize, Serialize)]
+pub(crate) struct SessionToken {
+    #[serde(
+        serialize_with = "serialize_object_id",
+        deserialize_with = "deserialize_object_id"
+    )]
+    pub(crate) object_id: ObjectId,
```
lcian, line 32: think if we need this

objectstore-service/src/resumable.rs
```diff
@@ -0,0 +1,356 @@
...
+
+    /// Encrypts a structured Resumable Upload session token.
+    pub(crate) fn encrypt(&self, session: SessionToken) -> Result<EncryptedSessionToken> {
+        let key_id = self.active_key_id.as_bytes();
+
+        let mut header = Vec::with_capacity(1 + key_id.len());
+        header.push(key_id.len() as u8);
+        header.extend_from_slice(key_id);
+
+        let mut ciphertext = serde_json::to_vec(&session).context(
```
jan-auer, line 126: JSON escapes certain character sequences in strings. It would be good to explore other serialization formats like `bincode`. 

At the same time we have to consider forward compatibility (if we keep the structured session token). JSON has named keys, so it's easy to add new ones, while `bincode` will break if the struct changes.

objectstore-service/src/resumable.rs
```diff
@@ -0,0 +1,356 @@
...
+}
+
+impl Encryptor {
+    /// Constructs an encryptor with a fresh process-local AES-256 key.
+    ///
+    /// Tokens encrypted with this key cannot be decrypted after a service restart. Configure a
+    /// persistent keyring with [`Self::new`] when resumable sessions must survive restarts.
+    ///
+    /// Returns an error if secure random key generation fails.
+    pub fn ephemeral() -> anyhow::Result<Self> {
```
jan-auer, line 70: Two options here:
 - Let's try to make this infallible by using an infallible random generator and use a variation of `new` that always succeeds.
 - If this is not possible, we could maybe lift the construction of the ephemeral encryptor out of the service constructor and pass it during startup?

objectstore-service/src/backend/gcs.rs
```diff
@@ -461,6 +463,48 @@ fn insert_gcs_meta_header(
...
+    fn from_token(token: &BackendToken, endpoint: &Url) -> Result<Self> {
+        let (total_length, session_uri) = token
+            .split_once('.')
+            .ok_or(ErrorKind::UnknownUploadSession)?;
+        let total_length = total_length
+            .parse()
+            .map_err(|_| ErrorKind::UnknownUploadSession)?;
+        let session_uri = Url::parse(session_uri).map_err(|_| ErrorKind::UnknownUploadSession)?;
+        let session = Self::new(session_uri, total_length);
+        if session.session_uri.origin() != endpoint.origin() {
```
lcian, line 501: remove this check

objectstore-service/src/backend/gcs.rs
```diff
@@ -461,6 +463,48 @@ fn insert_gcs_meta_header(
...
+impl ResumableUpload {
+    fn new(session_uri: Url, total_length: u64) -> Self {
+        Self {
+            session_uri,
+            total_length,
+        }
+    }
+
+    fn into_token(self) -> BackendToken {
+        format!("{}.{}", self.total_length, self.session_uri)
```
jan-auer, line 489: As a follow-up, let's investigate if we can move the responsibility of stringifying up into the encoder / the service. This all depends on how we can send a typed backend token generically through the tiered backend.

objectstore-service/src/backend/gcs.rs
```diff
@@ -740,6 +803,96 @@ impl fmt::Debug for GcsBackend {
...
+    let end = value
+        .strip_prefix("bytes=0-")
+        .filter(|end| !end.is_empty() && end.bytes().all(|byte| byte.is_ascii_digit()))
+        .and_then(|end| end.parse::<u64>().ok())
+        .ok_or_else(|| {
+            Error::new(
+                ErrorKind::BackendFailure,
+                "GCS: malformed resumable Range header",
+            )
+        })?;
```
jan-auer, line 817: The source for the parse error can be useful. Let's remove the filter, always try to parse, and then use `.context(ErrorKind::BackendFailure, "...")`

objectstore-service/src/backend/gcs.rs
```diff
@@ -919,6 +1068,206 @@ impl Backend for GcsBackend {
...
+            .checked_add(content_length)
+            .filter(|end| *end <= session.total_length)
+            .ok_or(ErrorKind::ChunkExceedsUploadLength {
+                offset,
+                content_length,
+                upload_length: session.total_length,
+            })?;
+
+        let content_range = match content_length {
+            // If `total_length` of this upload is 0, the only way to complete it is to
```
jan-auer, line 1169: Can we reject such uploads during session creation? While techincally possible, we probably do not want to allow uploading 0-byte files. 

This is an edge-case we haven't considered before and I do see legitimate use cases for empty attachments (e.g. a 0-byte log file), but in any case it makes to sense to start a resumable upload then.

objectstore-service/src/backend/gcs.rs
```diff
@@ -919,6 +1068,206 @@ impl Backend for GcsBackend {
...
+            .request(Method::PUT, session.session_uri.as_str())
+            .await?
+            .header(header::CONTENT_LENGTH, content_length)
+            .header(header::CONTENT_RANGE, content_range)
+            .body(Body::wrap_stream(stream))
+            .send_traced()
+            .await
+            .reqwest_context("GCS: upload resumable chunk")?;
+
+        range_response_to_upload_progress(&session, response)
```
lcian, line 1186: use an enum for `completed`

objectstore-service/src/backend/gcs.rs
```diff
@@ -919,6 +1068,206 @@ impl Backend for GcsBackend {
...
+            .await
```
jan-auer, line 1187: Can we use `.awat?` here?

objectstore-server/src/config.rs
```diff
@@ -629,6 +632,61 @@ pub struct Service {
...
+    /// their sessions may still be active; removing a key intentionally invalidates those sessions.
+    ///
+    /// ```yaml
+    /// service:
+    ///   resumable_token_encryption:
+    ///     active_key_id: v1
+    ///     key_files:
+    ///       v1: /var/run/secrets/objectstore/resumable-upload-v1
+    /// ```
+    pub resumable_token_encryption: Option<ResumableTokenEncryptionConfig>,
```
jan-auer, line 650: I'd suggest to make this configuration independent of resumable uploads. This is a key that can be used for any symmetric encryption business, which includes but does not have to be limited to resumable uploads.

objectstore-server/src/config.rs
```diff
@@ -629,6 +632,61 @@ pub struct Service {
...
+}
+
+impl Service {
+    /// Loads and validates the configured resumable token encryption keys.
+    pub(crate) fn resumable_token_encryption(&self) -> Result<Option<Encryptor>> {
```
jan-auer, line 655: As per #599, there's a way to include file contents directly via config. This is also what the new kafka config uses. Let's check if we can use this instead.

objectstore-service/src/resumable.rs
lcian, line 1: check if theres a higher level lib to do enc/decryption

objectstore-types/src/resumable.rs
```diff
@@ -35,48 +37,75 @@ const OFFSET_WILDCARD: &str = "*";
...
+        URL_SAFE_NO_PAD.encode(&self.0)
     }
 }
 
-impl From<String> for SessionToken {
-    fn from(token: String) -> Self {
-        Self(token)
+impl fmt::Debug for SessionToken {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        f.write_str("SessionToken([redacted])")
```
jan-auer, line 81:
```suggestion
        f.write_str("SessionToken")
```

objectstore-types/src/resumable.rs
```diff
@@ -35,48 +37,75 @@ const OFFSET_WILDCARD: &str = "*";
...
 
-impl fmt::Display for SessionToken {
-    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
-        self.0.fmt(f)
+impl<'de> Deserialize<'de> for SessionToken {
+    fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
+    where
+        D: Deserializer<'de>,
+    {
+        let encoded = String::deserialize(deserializer)?;
```
jan-auer, line 99: Consider deserializing into a `Cow` here, which can borrow and doesn't allocate.

objectstore-server/src/resumable.rs
```diff
@@ -177,19 +175,8 @@ where
...
-    if URL_SAFE_NO_PAD.encode(&bytes) != encoded {
-        return Err(ApiError::client(
-            "session token must use unpadded base64 URL encoding",
-        ));
-    }
-
-    String::from_utf8(bytes)
-        .map(SessionToken::from)
-        .map_err(|error| ApiError::map_client("session token is not valid UTF-8", error))
+    SessionToken::from_base64url(encoded)
```
jan-auer, line 178: Function can be removed and inlined.

objectstore-service/src/backend/gcs.rs
```diff
@@ -461,6 +463,45 @@ fn insert_gcs_meta_header(
...
+        let (total_length, session_uri) = token
+            .split_once('.')
+            .ok_or(ErrorKind::UnknownUploadSession)?;
+        let total_length = total_length
+            .parse()
+            .map_err(|_| ErrorKind::UnknownUploadSession)?;
+        let session_uri = Url::parse(session_uri).map_err(|_| ErrorKind::UnknownUploadSession)?;
+        let session = Self::new(session_uri, total_length);
+        Ok(session)
+    }
```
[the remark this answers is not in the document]
lcian: Intended. The `Location` returned from the google apis is not necessarily the same base URL we're using when we issue the request (it's just a opaque url, it's not documented otherwise).

objectstore-service/src/resumable.rs
lcian: > If this is not possible, we could maybe lift the construction of the ephemeral encryptor out of the service constructor and pass it during startup?

Done. It makes `StorageService::new` infallible again, the rest remains the same.

objectstore-server/src/config.rs
lcian: I'm going to do this in a follow-up, along with making the `Encryptor` generic or at least more general.

objectstore-server/src/config.rs
lcian: Changed it to use `serde_vars` (the lib introduced in https://github.com/getsentry/objectstore/pull/599).
To do this without boilerplate, I've added dependencies to `serde_bytes` (already a dependency of `serde_vars` itself) and `serde_with` (https://crates.io/crates/serde_with).

objectstore-service/src/backend/gcs.rs
lcian: Tracking this in my personal TODOs as a follow-up, along with the changes to the Encryptor.

objectstore-service/src/backend/gcs.rs
```diff
@@ -919,6 +1065,199 @@ impl Backend for GcsBackend {
...
+                .reqwest_context("GCS: query resumable upload")?;
+
+            range_response_to_upload_progress(&session, response)
+                .await
+                // If we get a completed object back it means that a previous `put_chunk` request
+                // already observed it and wrote to the `change_stream`. If we were to write to it
+                // again we would double count.
+                .map(|(progress, _)| progress)
+        })
+        .await
```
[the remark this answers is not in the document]
lcian: Fixed in f65fe9a. The recovery offset query now reports the completed object, with an emulator-backed regression test for the truncated final response.

objectstore-service/src/backend/gcs.rs
```diff
@@ -919,6 +1065,199 @@ impl Backend for GcsBackend {
...
```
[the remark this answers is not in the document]
lcian: Fixed in f65fe9a. The recovery offset query now reports the completed object, with an emulator-backed regression test for the truncated final response.

objectstore-service/src/backend/gcs.rs
```diff
@@ -919,6 +1065,209 @@ impl Backend for GcsBackend {
...
+                            id,
+                            stored_size,
+                            object.metadata_size(),
+                            object.custom_time,
+                        );
+                    }
+                    progress
+                })
+        })
+        .await
```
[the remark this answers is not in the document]
lcian: Not changing: this is intentional at-least-once state reporting. A completed status query can be the only confirmation after GCS commits a final chunk but its response cannot be parsed; suppressing it would undercount. Records retain a stable object-derived ID.

objectstore-service/src/resumable.rs
lcian: I've checked a few libs but they don't help _a lot_ with what we're doing as most of it is configuration, loading and validating the keys, etc.

lcian: Also while they might have a nice API, they implement other standards, are not nearly as widespread as `ring`, haven't been independently audited, etc.

objectstore-service/src/backend/gcs.rs
```diff
@@ -919,6 +1079,209 @@ impl Backend for GcsBackend {
...
+                        .await?;
+
+                    if response.status() != StatusCode::OK {
+                        let status = response.status();
+                        response.drain_body().await;
+                        return Err(Error::new(
+                            ErrorKind::BackendFailure,
+                            format!("GCS: unexpected resumable session creation status {status}"),
+                        ));
+                    }
```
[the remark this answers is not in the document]
lcian: JSON API replies 200 OK.

objectstore-service/src/backend/gcs.rs
lcian: Done. In addition we're using `NonZeroU64` throughout the backends to enforce this via types.

objectstore-service/src/backend/gcs.rs
```diff
@@ -919,6 +1080,196 @@ impl Backend for GcsBackend {
...
+                        response
+                            .check_error("GCS: cancel resumable upload")
+                            .await?
+                            .drain_body()
+                            .await;
+                        Err(Error::new(
+                            ErrorKind::BackendFailure,
+                            "GCS: unexpected resumable cancellation status",
+                        ))
+                    }
```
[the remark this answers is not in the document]
lcian: GCS returns 499 on successful cancels.

objectstore-service/src/resumable.rs
lcian: I would keep the JSON for now. The overhead should not be that high, normally.

objectstore-service/src/resumable.rs
lcian: We need it to check that the passed in `id` matches what's embedded in the token (the `id` that the token was originally created for).
It makes sense to centralize this logic in the `service`, as otherwise every backend would need to do the very same thing on its own.

objectstore-service/src/resumable.rs
```diff
@@ -0,0 +1,398 @@
...
+    }
+
+    /// Encrypts a structured Resumable Upload session token.
+    ///
+    /// Envelope format:
+    /// ```text
+    /// +------------++------------------------+------------------------+---------------------------+--------------------+
+    /// |            || Header                 | Nonce                  | Ciphertext                | Authentication tag |
+    /// +------------++------------------------+------------------------+---------------------------+--------------------+
+    /// | Contains   || ID length + key ID     | random                 | serde_json(token)         | AES-GCM verifier   |
```
lcian: Add 1 B version.

objectstore-service/src/resumable.rs
```diff
@@ -0,0 +1,398 @@
...
+    /// | Protection || public (authenticated) | public (authenticated) | encrypted + authenticated | public; checked    |
+    /// | Encoding   || 1 B length + UTF-8     | 12 B                   | variable                  | 16 B               |
```
lcian: Look if there's a standardized format instead of making up our own.

jan-auer (approved):

objectstore-server/src/config.rs
```diff
@@ -629,6 +633,58 @@ pub struct Service {
...
+#[derive(Clone, Deserialize, Serialize)]
+pub struct ResumableTokenEncryptionConfig {
+    /// Key used to encrypt newly created sessions.
+    pub active_key_id: String,
+    /// Raw, exactly 32-byte AES-256 keys, indexed by rotation ID.
+    ///
+    /// File-backed secrets should use `${file:PATH}` so they are loaded during configuration
+    /// deserialization.
+    #[serde(default)]
+    #[serde_as(as = "BTreeMap<_, Bytes>")]
```
lcian: Check if we can use `bytes::Bytes` directly.

objectstore-service/src/resumable.rs
lcian: I've reviewed some approaches.
Ours is practically equivalent to JWE Compact (https://datatracker.ietf.org/doc/html/rfc7516#page-10).
The only difference is that JWE compact applies b64url encoding to each component and uses a JSON header that contains algorithm, encryption scheme and key id, while we use a custom binary format for the header.

Our encoding has the same features, has less overhead as we don't JSON-encode the header, and I think it's nicer to leave the transport encoding (b64url) to the server instead of relying on the envelope being already correctly encoded, so I'm leaving this as is (after adding the version byte at the start).

objectstore-service/src/backend/gcs.rs
```diff
@@ -919,6 +1081,196 @@ impl Backend for GcsBackend {
...
+                    _ => {
+                        response
+                            .check_error("canceling a GCS resumable upload")
+                            .await?
+                            .drain_body()
+                            .await;
+                        Err(Error::new(
+                            ErrorKind::BackendFailure,
+                            "unexpected GCS resumable upload cancellation status",
+                        ))
```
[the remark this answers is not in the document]
lcian: The implementation correctly matches 499 first via CLIENT_CLOSED_REQUEST_STATUS, then treats 404 and 410 as idempotent success (objectstore-service/src/backend/gcs.rs:1246). The nearby 204 behavior belongs to the separate XML API multipart implementation (objectstore-service/src/backend/gcs.rs:1540).

Although check_error would pass a hypothetical 204 through, the subsequent BackendFailure is intentional strict status validation: 204 is unexpected for this JSON API operation. Accepting it could conceal a protocol or endpoint mismatch.

PR #611, opened 2026-09-04
fix(bigtable): Make RPC timeout configurable

jan-auer:
Objectstore currently allows each Bigtable RPC attempt to run for 10 seconds. With two retries, this can exceed Envoy’s 15-second per-try timeout.

Add a configurable Bigtable `rpc_timeout` with a default of 2 seconds. The timeout is intentionally short so stalled RPCs fail faster and Objectstore can retry within Envoy’s upstream timeout budget.

[getsentry/ops#23284](https://github.com/getsentry/ops/pull/23284) makes the corresponding Envoy route timeout budgets explicit.

Ref FS-502


lcian (approved):
Good change.

PR #614, opened 2026-09-07
fix(service): Move TTI renewal to the background

jan-auer:
Backend GET and HEAD operations currently renew TTI deadlines inline. These rewrites can race with concurrent tiered-storage commits and overwrite newer state.

This change makes backend reads side-effect-free. The service schedules bounded, best-effort renewals after successful reads. It adds extend-only conditional expiry updates for each backend and renews tiered objects blob-first before updating the matching redirect.

Ref FS-405

### DIFF
--- a/objectstore-service/src/backend/bigtable.rs
+++ b/objectstore-service/src/backend/bigtable.rs
@@ -45,7 +45,7 @@ use tracing::Instrument;
 
 use crate::backend::common::{
     Backend, DeleteResponse, GetResponse, HighVolumeBackend, MetadataResponse, PutResponse,
-    TieredGet, TieredMetadata, TieredWrite, Tombstone,
+    TieredGet, TieredMetadata, TieredUpdate, TieredWrite, Tombstone,
 };
 use crate::change_stream::{
     ChangeStream, ChangeStreamFactory, CostTrackerStreamConfig, flush_change_stream,
@@ -1061,7 +1061,13 @@ impl Backend for BigTableBackend {
     }
 
     async fn set_expiry(&self, id: &ObjectId, expire_at: SystemTime) -> Result<bool> {
-        <Self as HighVolumeBackend>::set_expiry_if_matches(self, id, expire_at, None).await
+        <Self as HighVolumeBackend>::compare_and_update(
+            self,
+            id,
+            None,
+            TieredUpdate::SetExpiry(expire_at),
+        )
+        .await
     }
 
     #[tracing::instrument(level = "debug", skip(self))]
@@ -1206,12 +1212,14 @@ impl HighVolumeBackend for BigTableBackend {
     }
 
     #[tracing::instrument(level = "debug", skip(self))]
-    async fn set_expiry_if_matches(
+    async fn compare_and_update(
         &self,
         id: &ObjectId,
-        expire_at: SystemTime,
-        expected_target: Option<&ObjectId>,
+        current: Option<&ObjectId>,
+        update: TieredUpdate,
     ) -> Result<bool> {
+        let TieredUpdate::SetExpiry(expire_at) = update;
+        let expected_target = current;
         let expire_at = super::common::normalize_expiry(expire_at)?;
         let path = id.as_storage_path().to_string().into_bytes();
 
@@ -1939,17 +1947,21 @@ mod tests {
         let later = old_expiry + Duration::from_hours(2);
         assert!(
             !backend
-                .set_expiry_if_matches(&id, later, Some(&wrong_target))
+                .compare_and_update(&id, Some(&wrong_target), TieredUpdate::SetExpiry(later),)
                 .await?
         );
         assert!(
             backend
-                .set_expiry_if_matches(&id, later, Some(&target))
+                .compare_and_update(&id, Some(&target), TieredUpdate::SetExpiry(later))
                 .await?
         );
         assert!(
             backend
-                .set_expiry_if_matches(&id, old_expiry + Duration::from_mins(30), Some(&target),)
+                .compare_and_update(
+                    &id,
+                    Some(&target),
+                    TieredUpdate::SetExpiry(old_expiry + Duration::from_mins(30)),
+                )
                 .await?
         );
         let TieredMetadata::Tombstone(tombstone) = backend.get_tiered_metadata(&id).await? else {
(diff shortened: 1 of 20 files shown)


objectstore-service/src/backend/common.rs
```diff
@@ -270,6 +281,22 @@ pub trait HighVolumeBackend: Backend {
...
     ) -> Result<bool>;
+
+    /// Atomically updates an existing row if its kind and redirect target match.
+    ///
+    /// `current = None` requires a live inline object. `Some(target)` requires
+    /// a live redirect to exactly that target. Updates never authorize creation
+    /// of an absent row.
+    ///
+    /// Returns `true` when the update was applied or its requested state was
+    /// already satisfied. Returns `false` for an absent, expired, manual-policy,
```
lcian:
```suggestion
    /// already satisfied. Returns `false` for an absent, expired,
```
Given that this performs a generic `TieredUpdate`, the fact that we fail the op when the policy is manual is probably more suited to rather be documented on `TieredUpdate::SetExpiry`.

objectstore-service/src/backend/gcs.rs
```diff
@@ -768,9 +750,12 @@ impl GcsBackend {
                 .await
                 .reqwest_context("updating GCS custom time")?;
 
-            // Bumping TTI is opportunistic. A concurrent metadata writer won the CAS race, so
-            // leave its update intact and let a future read evaluate the TTI again.
-            if response.status() == StatusCode::PRECONDITION_FAILED {
+            // A concurrent metadata writer won the CAS race. Leave its update
+            // intact; automatic renewal can be retried by a later read.
```
lcian:
```suggestion
            // intact.
```

objectstore-service/src/backend/gcs.rs
```diff
@@ -976,12 +961,15 @@ impl Backend for GcsBackend {
...
+        let Some(gcs_metadata) = self.get_gcs_metadata(&object_url).await? else {
             return Ok(None);
         };
 
         let mut download_url = object_url;
-        download_url.query_pairs_mut().append_pair("alt", "media");
+        download_url
+            .query_pairs_mut()
+            .append_pair("alt", "media")
+            .append_pair("ifGenerationMatch", &gcs_metadata.generation);
```
lcian: Why only `ifGenerationMatch` but not `ifMetagenerationMatch` I wonder?

objectstore-service/src/backend/counting.rs
```diff
@@ -89,6 +90,12 @@ impl Backend for CountingBackend {
         self.inner.get_metadata(id).await
     }
 
+    async fn set_expiry(&self, id: &ObjectId, expire_at: SystemTime) -> Result<bool> {
+        // Renewal is an implementation detail of the original client read and
+        // must not add another client-operation COGS unit.
```
lcian: Right now this is true, but when we change the API to require explicit bumps, we'll have to change this.

objectstore-service/src/backend/bigtable.rs
```diff
@@ -461,6 +461,78 @@ fn optional_target_predicate(target: &ObjectId, own_id: &ObjectId) -> MutatePred
...
+            ],
+        })),
+    })
+}
+
+/// Matches an inline row whose metadata cell has the observed expiry timestamp.
+///
+/// Millisecond expiry timestamps are not unique revisions. A replacement inline
+/// object with the same row kind and expiry can therefore still be overwritten
+/// by this rewrite. A dedicated persisted revision token is deferred.
```
lcian:
```suggestion
```
It seems that this is not relevant.

objectstore-service/src/background.rs
```diff
@@ -0,0 +1,339 @@
...
+}
+
+/// Schedules bounded, deduplicated expiry renewals in the background.
+#[derive(Debug)]
+pub struct RenewalScheduler {
+    inner: Arc<RenewalSchedulerInner>,
+    worker: Option<RenewalWorker>,
+}
+
+impl Clone for RenewalScheduler {
```
lcian, line 166: Why is this needed?

objectstore-service/src/backend/gcs.rs
jan-auer: We read metadata first and now need a matching payload. If only metadata changes in the meanwhile, the read is still valid.

objectstore-service/src/backend/counting.rs
jan-auer: That is true. Since we will only call set_expiry conditionally, I think we can change this right away.

objectstore-service/src/background.rs
jan-auer: The scheduler is like a handle to the worker. Instead of externally `Arc`'ing it, we clone it directly for the background metrics emitter and for stream executor.

objectstore-service/src/background.rs
```diff
@@ -0,0 +1,321 @@
...
+    pub fn schedule(&self, id: ObjectId, expire_at: SystemTime) {
+        let pending = Arc::clone(&self.inner.pending);
+        if !pending.pin().insert(id.clone()) {
+            objectstore_metrics::count!(
+                "service.expiry_renewal",
+                outcome = "skipped",
+                reason = "duplicate"
+            );
+            return;
+        }
```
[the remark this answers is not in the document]
jan-auer: Non-issue. If the queue is full, the other scheduling attempt wouldn't work either.

objectstore-service/src/background.rs
```diff
@@ -0,0 +1,321 @@
...
+            objectstore_metrics::count!("service.expiry_renewal", outcome = "dropped");
+            return;
+        }
+
+        let permit = loop {
+            match self.concurrency.acquire_bulk().await {
+                Ok(permit) => break permit,
+                Err(error) if error.kind() == ErrorKind::AtCapacity => continue,
+                Err(error) => {
+                    objectstore_metrics::count!(
```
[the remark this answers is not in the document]
jan-auer: That's the point of the concurrency permits: Processing should wait for an available permit before proceeding. There's no busy loop, because the acquire function actually waits and suspends.

objectstore-service/src/backend/gcs.rs
lcian: Ok fair.

objectstore-service/src/background.rs
lcian: Perhaps it could be called something like `RenewalQueue` rather than scheduler.

lcian (approved):
Looks good to me.

PR #615, opened 2026-09-08
ref(service): Generalize encryption util

lcian:
Authenticated encryption is now moved from `resumable.rs` to a reusable `objectstore-service::encryption::Cipher` struct whose `encrypt` and `decrypt` methods work with any serde type.

`resumable_token_encryption` has been renamed to `cipher/encryption` throughout, so configuration now happens as follows:
```yaml
service:
  encryption: ...
```


lcian:
bugbot run
@sentry review

objectstore-service/src/encryption.rs
```diff
@@ -0,0 +1,299 @@
...
+        envelope.extend_from_slice(&nonce);
+        envelope.extend_from_slice(&ciphertext);
+        Ok(envelope)
+    }
+
+    /// Decrypts and deserializes an encrypted value.
+    ///
+    /// Returns `None` if the envelope is malformed, cannot be authenticated, or does not contain
+    /// the requested type.
+    pub(crate) fn decrypt<T>(&self, envelope: &[u8]) -> Option<T>
```
lcian: I'm not sure if this is right or if we should rather return a `Result` here.
`Option` can still be mapped externally and has the advantage of not exposing the source `Error` in case anything fails, which could provide some info to an attacker.
So perhaps `Option` is the right fit here.

objectstore-server/src/config.rs
```diff
@@ -634,28 +634,32 @@ pub struct Service {
...
 
-    /// Persistent encryption keys for resumable-upload session tokens returned to clients.
+    /// Persistent encryption keys.
     ///
-    /// Tokens are always encrypted. When this is absent, Objectstore generates a fresh in-memory
-    /// AES-256 key at startup, so resumable sessions become invalid after a restart. Configure a
-    /// persistent keyring for sessions that must survive restarts. Keep old keys configured while
-    /// their sessions may still be active; removing a key intentionally invalidates those sessions.
-    /// Values must be raw AES-256 key bytes.
+    /// This is used to instantiate [`Cipher`] which can encrypt/decrypt arbitrary payloads for
```
jan-auer: _nit_: This should better be directed at an operator rather than engineer, hence describe what functionality this is used for and how it changes objectstore's behavior, rather than the code-level constructs.

We can put relevant code comments onto the `cipher` method.

objectstore-server/src/config.rs
```diff
@@ -634,28 +634,32 @@ pub struct Service {
...
+    /// data that should remain confidential when crossing the service boundary.
+    /// Currently, this is only used for session tokens of the Resumable Uploads API.
+    ///
+    /// When this config is absent, Objectstore generates a fresh in-memory key at startup.
+    /// This might not work for requests that need to survive a deployment or multi-instance
```
jan-auer: This _will_ not work for certain.

objectstore-service/src/encryption.rs
jan-auer: It is useful to keep the source error here for internal error tracking and rather restrict sources in the public API.

objectstore-server/src/config.rs
```diff
@@ -665,26 +669,26 @@ impl Service {
...
+pub struct EncryptionConfig {
+    /// Key used to encrypt new values.
     pub active_key_id: String,
-    /// Raw, exactly 32-byte AES-256 keys, indexed by rotation ID.
+    /// Exactly 32 raw key bytes, indexed by ID.
     ///
     /// File-backed secrets should use `${file:PATH}` so they are loaded during configuration
     /// deserialization.
     #[serde(default)]
     pub keys: BTreeMap<String, Bytes>,
```
jan-auer: It's nice that this works with Bytes, but if we need exactly 32 raw key bytes we can:
 - Implement our own newtype that performs the deserialization
 - Consider to store the bytes inline as `[u8; 32]`.

jan-auer (approved):

objectstore-server/src/config.rs
lcian: For resumable uploads it will not work, for other things that we use this in the future, it might...

objectstore-server/src/config.rs
lcian: Good point!

lcian:
bugbot run
@sentry review

PR #616, opened 2026-09-08
ref(service): Standardize backend session token codecs

lcian:
Backends can now define a sized session-token representation and centralize its codec while the
object-safe resumable API continues passing an opaque `BackendToken`. GCS owns the first typed
implementation and decodes tokens at each resumable operation boundary; decorators and backends
without resumable support continue forwarding or rejecting opaque tokens as before.

`BackendToken` is now a serde-transparent newtype, so its JSON representation, encrypted session
envelope, and GCS `total_length.session_uri` encoding remain wire-compatible. This PR is stacked on
#615, which provides the generalized encryption boundary used by these tokens.

A disadvantage of this approach is that the backend-specific types need to become `pub`, as they become part of the `Backend` API which is `pub` itself.

### DIFF
--- a/objectstore-service/src/backend/bigtable.rs
+++ b/objectstore-service/src/backend/bigtable.rs
@@ -987,6 +987,8 @@ impl BigTableBackend {
 
 #[async_trait::async_trait]
 impl Backend for BigTableBackend {
+    type SessionToken = ();
+
     fn name(&self) -> &'static str {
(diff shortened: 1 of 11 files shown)


lcian:
bugbot run
@sentry review

lcian:
I'm not sure if this change is beneficial, wdyt @jan-auer?

PR #622, opened 2026-09-10
feat(rust-client): Add Resumable Uploads (low-level) API

lcian:
Adds a feature-gated low-level `resumable-upload-api` to the Rust client.
We discourage the use of this API in docs.

A new `ResumableUploadError` variant is introduced that maps known failures of the resumable upload endpoints to specific variants.
The API still needs to return the more generic `crate::Result` to cover other kinds of errors (metadata, auth, rate limits, etc.).

A minor change to the server has been made to return 404 for unknown sessions instead of 400, so that we can reliably classify such failures.

The PR also adds all-features doctest coverage, which was previously missing, as `cargo llvm-cov` would need the `--doctetsts` flag to cover those, which itself requires nightly.

Close FS-472


lcian:
bugbot run
@sentry review

lcian:
bugbot run
@sentry review

clients/rust/src/client.rs
```diff
@@ -583,6 +583,20 @@ impl Session {
         self.prepare_builder(builder)
     }
 
+    #[cfg(feature = "resumable-upload-api")]
```
lcian, line 586: I called it `resumable-upload-api` to clearly distinguish it from the _capability_ to use resumable uploads at all, which we eventually want to use under the hood for regular puts automatically.
Calling it `resumable-upload` or `resumable-uploads` could mislead users into thinking they need that flag for the client to be able to use resumable _at all_.

clients/rust/src/resumable.rs
```diff
@@ -0,0 +1,392 @@
...
+        &self.key
+    }
+
+    /// Returns the session token bound to this upload.
+    pub fn token(&self) -> &SessionToken {
+        &self.token
+    }
+
+    /// Builds a request for the server's authoritative upload progress.
+    pub fn progress(&self) -> UploadProgressBuilder {
```
lcian, line 118: Here we could very well make this `pub async fn progress` directly, but I wanted to stick to our existing pattern that requires you to explicitly call `send` whenever an operation triggers a network request.
Plus, it gives us the possibility to extend these builders with additional params without a breaking change.

clients/rust/src/client.rs
jan-auer: While this is being developed, it's a good idea to gate this behind a feature flag. Afterwards, especially if we have resumable uploads inside normal `put`, we should either:
 - Make this a `resumable-upload` feature that gates both
 - (preferred) Remove the feature entirely. There are no dependencies that it is gating and compilation is not much faster without.

clients/rust/src/client.rs
```diff
@@ -583,6 +583,20 @@ impl Session {
...
+    pub(crate) fn resumable_request(
+        &self,
+        method: reqwest::Method,
+        object_key: &str,
+        query_pair: (&str, &str),
```
jan-auer, line 591: This method doesn't really create a resumable upload request. It's just a regular request with an arbitrary query pair added to the URL. I think we should either type this with the kinds of different requests we support (`session` vs `uploadType`), or choose a more generic name.

`RequestBuilder` also has a `query` method that allows you to set a typed query string. I think the cleanest version would be to construct a regular request and then use `.query()` to set the string outside.

clients/rust/README.md
```diff
@@ -133,6 +133,121 @@ session.put("payload")
...
+### Resumable Upload API
+
+> **Feature flag required:** Enable `resumable-upload-api` to use this API.
+>
+> ```toml
+> objectstore-client = { version = "...", features = ["resumable-upload-api"] }
+> ```
+
+The resumable upload API allows you to upload an object across multiple requests.
+It's suitable for use for particularly large objects, where restarting an upload
```
jan-auer, line 145:
```suggestion
It is particularly suitable for large objects, where restarting an upload
```

clients/rust/README.md
```diff
@@ -133,6 +133,121 @@ session.put("payload")
...
+It's recommended to always try to upload the whole object in a single request if
+possible, as that's always the more efficient approach.
+If the request fails midway, it will be possible to resume it from the persisted offset.
+
+**Note:** This feature flag exposes a low-level API that maps directly to the server API and
+requires appropriate manual handling of different states and error scenarios.
+Therefore, this API should only be used for advanced use cases that demand it.
+In a future release of `objectstore-client`, the resumable uploads API will be used
+internally for eligible `put` calls without the need for this feature flag or direct
+interaction with this API.
```
jan-auer, line 156: This seems alarming without detail, but also doesn't offer an alternative. For example "different states and error scenarios" is vague.

I would skip this note for now and add a simple callout later when we have this integrated into regular puts: "Note that the objectstore client uses resumable uploads automatically for large files. See [...]."

clients/rust/README.md
```diff
@@ -133,6 +133,121 @@ session.put("payload")
...
+
+**Important:** resumable uploads do not automatically compress chunk contents. The `compression`
+setting only records how the object is encoded; the caller must compress the payload accordingly.
+The object length and all offsets refer to the bytes after compression.
+
+```rust,no_run
+#[cfg(feature = "resumable-upload-api")]
+mod example {
```
jan-auer, line 164: This can be simplified without a module and I'd hide the async function wrapper to keep this example focused on the steps required to perform a successful upload; stripping out all optional things (like setting content type or compression).

clients/rust/README.md
```diff
@@ -133,6 +133,121 @@ session.put("payload")
...
+                .send()
+                .await;
+
+            offset = match result {
+                Ok(UploadProgress::Complete) => return Ok(()),
+
+                Err(Error::ResumableUpload(error @ ResumableUploadError::Gone))
+                | Err(Error::ResumableUpload(
+                    error @ ResumableUploadError::NotFound,
+                )) => {
```
jan-auer, line 211: This example shows that the error surface of resumable uploads is complex. On first glance, a few things we can improve:

- It's required to match specific sub-variants of the `ResumableUpload` error variant to understand that this is a terminal error. We should centralize this in a helper method on Error; or create a helper for the inverse (can still retry)
- Same for offset mismatches: The user now has to combine an `Ok` response (incomplete) and `Err` response (offset mismatch). It is technically correct to classify these two differently, but for internal and external use I believe those should better be folded into the same `UploadProgress` variant.

clients/rust/src/resumable.rs
```diff
@@ -0,0 +1,382 @@
...
+    ///
+    /// Fall back to a regular [`Session::put`].
+    #[error("resumable upload was declined")]
+    Declined,
+}
+
+/// A handle bound to one resumable upload session.
+///
+/// See the [crate-level documentation](crate#resumable-upload-api) for more information and
+/// examples on using this API.
```
jan-auer, line 61: As a general thought: I think it's worth having simplified happy-path examples on the crate-level docs, but then have the more detailed or elaborate examples that maybe show edge cases or advanced use here (or even on the methods).

clients/rust/src/resumable.rs
```diff
@@ -0,0 +1,382 @@
...
+
+    /// Builds a request to write `chunk` starting at `offset`.
+    ///
+    /// `put_chunk` doesn't perform any automatic compression of the payload, so the caller is
+    /// responsible for applying compression to the entire payload beforehand and passing chunks
+    /// of the already compressed payload to this method.
+    ///
+    /// The returned progress contains the authoritative server offset, which should be used for
+    /// subsequent requests.
+    pub fn put_chunk(&self, offset: u64, chunk: impl Into<Bytes>) -> PutChunkBuilder {
```
jan-auer, line 132: What if we call this `put` to further emphasize the best practice of writing entire objects? The builder can still have a name distinct from `PutBuilder`.

clients/rust/src/resumable.rs
```diff
@@ -0,0 +1,382 @@
...
+        .get(HEADER_UPLOAD_OFFSET)?
+        .to_str()
+        .ok()?;
+    match value.parse().ok()? {
+        UploadOffset::At(offset) => Some(offset),
+        UploadOffset::Unknown => None,
+    }
+}
+
+fn parse_error_response(response: Response, operation: &str) -> Error {
```
jan-auer, line 360: We should consume the response body so connections can be reused.

clients/rust/src/client.rs
lcian: get_url then apply .query

clients/rust/README.md
lcian: Put this on the method, then we will remove the feature flag in the future, once this is stable.

clients/rust/README.md
lcian: `#![cfg(feature = "resumable-upload-api")]` try this

clients/rust/README.md
lcian: Try to put an `is_terminal` helper on the `ResumableUploadError`, or make the error variants themselves represent how to handle (retry/proceed/abort).

lcian: Or make the error variants themselves like retry/proceed/abort.

clients/rust/src/resumable.rs
lcian: Maybe make the example easier, without all the error handling, we already have "how to react" on the error variants themselves, and add more info to the methods if needed.

clients/rust/src/client.rs
lcian: We'll keep the flag while this is being developed, then remove it, and just make it clear that you should probably not use it. Then, automatic usage inside regular `put`s will be subject to a runtime flag, likely on the `session` builder or on the `put` builder itself.

clients/rust/src/resumable.rs
lcian: True. We actually need to do this everywhere in the client, not only for these APIs.

lcian: True.
We actually need to do this everywhere in the client, not only for these APIs, so I'll address all of them in an immediate follow-up once this is merged.

clients/rust/src/client.rs
lcian: We have the same pattern for multipart, so I'll change both in another immediate follow-up.

clients/rust/README.md
lcian: In https://github.com/getsentry/objectstore/pull/622/commits/e06aef6a9222192c225ed0d51538637f37101ade I removed `ResumableUploadError`.
Here's what we do for the variants that we had on that struct:
- `Declined` -> now `Session::create_upload` returns `Ok(None)` if the upload is declined (similar to 404 on get)
- `OffsetMismatch(offset)` -> now `Session::put` returns `Ok(offset)`, as it does when the request succeeds, just with a different offset.
- `Gone` / `NotFound` -> now represented by `objectstore_client::Error::ResumableUploadUnavailable` (note that this is on the general crate-level `Error`, as `ResumableUploadError` has been removed).

clients/rust/src/resumable.rs
lcian: Now we're still handling a couple errors where it makes sense in the `README.md` but with the new structure of errors it looks much nicer.

lcian:
bugbot run
@sentry review

clients/rust/src/resumable.rs
```diff
@@ -289,6 +265,26 @@ impl fmt::Debug for PutChunkBuilder {
...
+    /// Returns [`Error::ResumableUploadUnavailable`] when the session expired, was canceled, or
+    /// could not be found. The upload must be restarted with a new session in that case.
+    ///
+    /// ```rust,ignore
+    /// let offset = match upload.put(offset, chunk).send().await {
+    ///     Ok(UploadProgress::Complete) => return Ok(()),
+    ///     Ok(UploadProgress::Incomplete { offset }) => offset,
+    ///     Err(error @ Error::ResumableUploadUnavailable) => {
+    ///         // The caller must start the entire upload again with a new session.
+    ///         return Err(error);
```
jan-auer:
```suggestion
    ///     Err(Error::ResumableUploadUnavailable) => {
    ///         todo!("restart upload");
```

jan-auer (approved):

PR #623, opened 2026-09-10
fix(service): Coordinate local filesystem mutations per object

jan-auer:
Replace the backend-local write mutex introduced to `LocalFsBackend` in #614 with advisory filesystem locks so mutations of the same object coordinate across backend instances. The locks also have a smaller scope so that concurrent writes are still possible.

- The lock now only guards the region that mutates the object. Previously, we covered the entire file upload on PUTs, which reduced throughput. 
- Instead of per-object locks, the backend uses 65k permanent slots. The slot is determined by a stable `blake3` hash of the object key. This bounds the number of lock files on the file system while still minimizing contention. This approach particularly simplifies cleanup, since the backend does not have to worry about orphaning locks during crashes or leaving files behind for missing objects.
- To limit the number of open file handles and pending blocking tasks in tokio, we use an additional semaphore in front before acquiring a filesystem lock.


lcian (approved):
Looks good!
Should we mention somewhere explicitly that we don't support Windows? Although that might be pretty obvious already since we don't build/publish images for it.

objectstore-service/src/backend/local_fs.rs
```diff
@@ -534,6 +550,105 @@ impl MultipartUploadBackend for LocalFsBackend {
...
+            .await
+            .context(ErrorKind::Internal, "waiting for local-fs object lock")?
+            .context(ErrorKind::BackendFailure, "acquiring local-fs object lock")?;
+
+            if let Some(guard) = guard {
+                return Ok(guard);
+            }
+        }
+    }
+}
```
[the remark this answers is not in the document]
lcian: There are probably other places where this is an issue, so we can defer fixing this to  https://linear.app/getsentry/issue/FS-226/local-fs-prevent-path-traversal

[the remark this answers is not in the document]
jan-auer: Added to that issue.

objectstore-service/src/backend/local_fs.rs
```diff
@@ -5,14 +5,16 @@
...
-//! allowing an open file to be unlinked. Shared filesystems are supported only when locks propagate
-//! across the cluster, file identities are stable, pathname visibility is coherent, and rename is
-//! atomic. Locks are advisory, so all writers must cooperate. Each backend instance caps its
-//! contribution to blocking-pool occupancy from lock waiters.
-
+//! The first two bytes of the BLAKE3 hash of an object's storage path select one of 65,536
+//! permanent lock slots under `.locks/<first byte>/<second byte>` (lowercase hexadecimal).
+//! Files are created lazily; unrelated objects that hash to the same slot serialize mutations.
+//! Slot files must never be removed while writers are running. All cooperating writers must use
+//! the same mapping; stop writers using the old per-object layout before upgrading.
```
lcian:
```suggestion
//! Slot files must never be removed while writers are running. All cooperating writers must use
//! the same mapping.
```

lcian: (it doesnt let me suggest idk why) but basically the part after the `;` is irrelevant as the "old per-object layout" was never merged

jan-auer: good catch, thanks!

lcian (approved):

PR #626, opened 2026-09-14
fix(ci): Switch to quay.io for minio image

jan-auer:
Minio no longer maintains their docker registry, so we'll have to start building our own image or stop using minio for testing S3. In the interim, to unblock CI, we're using quay.io to obtain that image.


lcian (approved):

lcian:
Need to change also the CI job.

aldy505:
@jan-auer I would rather use this https://hub.docker.com/r/alpine/minio

PR #627, opened 2026-09-14
fix(service): Use second-precision timestamps across backends

jan-auer:
Add a shared `Timestamp` type for object creation, expiration, and expiry-related access times. Round fractional timestamps up to whole seconds to prevent precision drift between backends.

Preserve the existing metadata serialization format and provide an RFC3339 adapter for headers and GCS metadata. Keep metrics and other event timestamps at their current precision. Preserve raw Bigtable timestamps for exact conditional updates of existing rows.


lcian:
We could consider adding a rule such as the following to the `clippy.toml` of `objectstore-service`:
```toml
disallowed-types = [
    { path = "std::time::SystemTime", reason = "use objectstore_types::time::Timestamp" },
]
```

lcian (approved):
Looks good to me.

jan-auer:
I wouldn't do that. It's too likely that we still want to use SystemTime in places independent of expiry. It seems a bit too strict at the current time, but I'm happy to revisit.

PR #628, opened 2026-09-14
ref(rust-client): Use typed queries for resumable API

lcian:
Resumable upload requests now serialize `upload_type` and `session` through typed Serde query structs and Reqwest's query API.
There was another place where we do this which is the multipart API, but given that we're going to remove it I'm not gonna spend time on it.

Refs FS-514


lcian:
@jan-auer do you think it's worth it to use this internally also in the GCS backend? (as a follow-up)

jan-auer (approved):

PR #629, opened 2026-09-14
fix(rust-client): Drain discarded response bodies

lcian:
The Rust client now consumes response bodies on error and on successful paths that otherwise discard them, allowing the underlying HTTP connections to return cleanly to the connection pool.
Successful GET responses remain caller-owned streams, while JSON responses continue to be consumed by deserialization.

Special case: batch multipart responses remain non-drained, as that would require buffering the whole batch response at once due to a limitation in `multer`.

Refs FS-514


jan-auer (approved):
What a name :) 
Thanks!

PR #630, opened 2026-09-14
ref(service): Inject access time into storage operations

jan-auer:
Capture one timestamp per HTTP request and pass it through metadata creation, service calls, and backend expiry checks. Previously, backends sampled time independently. A slow request could therefore use different times for creation, expiry checks, and TTI renewal.

Use an explicit timestamp parameter to keep the interface small and dependencies visible. Endpoints extract the timestamp, and the auth wrapper forwards it like the other operation arguments. Reuse it across all batch items, tiered calls, and retries.

Background renewals and cleanup start separate operations and use fresh timestamps for liveness checks. This prevents time spent waiting in a queue from authorizing renewal of an expired object. Renewal deadlines still come from the original read, so queue delays do not extend retention.

Keep multipart modification timestamps and cleanup grace periods unchanged. They serve separate purposes from object access and expiry.


lcian (approved):
As we discussed, I don't foresee any cases where we would need to carry around more "request context" other than this timestamp, and I think carrying this in a task-local or similar implicit "context" would likely be error-prone. So, I'm fine with this approach.

PR #633, opened 2026-09-16
feat(server): Tag request durations with usecase

lcian:
Request duration distributions now include the usecase captured by Axum's matched route.
Requests without a `{usecase}` capture emit `usecase:none`, keeping the tag set consistent across every `server.requests.duration` data point -- iirc, it's awkward to query them in DataDog when the set of tags is different.


objectstore-server/src/web/metrics_body.rs
```diff
@@ -71,11 +79,16 @@ impl Drop for EmitMetricsGuard {
             BodyState::Completed(status) => status.as_u16(),
             BodyState::Errored => 500,
         };
-
+        let usecase = self
+            .usecase
+            .take()
+            .map(Into::into)
+            .unwrap_or(Cow::Borrowed("none"));
```
jan-auer: `map_or` would work well here:


```suggestion
            .map_or(Cow::Borrowed("none"), Cow::Owned);
```

objectstore-server/src/web/middleware.rs
```diff
@@ -107,10 +107,21 @@ pub async fn bind_sentry_body(request: Request, next: Next) -> Response {
 pub async fn emit_request_metrics(mut request: Request, next: Next) -> Response {
     let matched_path = request.extract_parts::<MatchedPath>().await;
     let route = matched_path.as_ref().map_or("unknown", |m| m.as_str());
+    let usecase = request
```
jan-auer: _nit_: this could be more readable in a helper function.

jan-auer (approved):
That's very useful!

PR #634, opened 2026-09-16
fix(service): Account for debounce in persisted TTI

lcian:
Time-to-idle deadlines now include the same `min(tti / 4, 24h)` window used to debounce renewals. 
A requested 7-day TTI therefore gets an 8-day backend deadline, preventing an access near the debounce boundary from receiving less than 7 days of remaining retention.


lcian:
This will be fixed by the explicit expiry bumps, which are not subject to the debounce,  (https://github.com/getsentry/objectstore/pull/637) and deprecation of TTI.

PR #636, opened 2026-09-16
feat(service): Support relative expiry targets

jan-auer:
Extend internal expiry updates to accept absolute deadlines or durations relative to object creation time. This contract will support the web API for bumping object expiry.

Each backend resolves creation-relative targets from metadata read by the mutation itself and returns the resolved requested deadline.

Tiered storage resolves the target against the long-term blob, then conditionally propagates that exact deadline to the redirect. This preserves existing update ordering and prevents a later blob deadline from over-extending the redirect.

Overflow clamps to the maximum supported timestamp. A creation-relative target without creation metadata is rejected. Existing absolute TTI renewal behavior remains unchanged.

Ref FS-438


objectstore-service/src/backend/gcs.rs
```diff
@@ -1072,18 +1072,22 @@ impl Backend for GcsBackend {
...
-            return Ok(false);
+            return Ok(None);
         };
         let Some(current_expiry) = object.custom_time else {
-            return Ok(false);
+            return Ok(None);
+        };
+        let Some(expire_at) = target.resolve(object.time_created.map(Rfc3339Timestamp::into_inner))
+        else {
+            return Ok(None);
```
lcian, line 1087: Perhaps `object.time_created.is_none()` should be an internal error, as we've added `time_created` several months ago before launching to production, so we can expect all objects to have it.

objectstore-service/src/backend/s3_compatible.rs
```diff
@@ -437,22 +437,26 @@ impl<T: TokenProvider> Backend for S3CompatibleBackend<T> {
...
         else {
-            return Ok(false);
+            return Ok(None);
         };
         let Some(current_expiry) = metadata.time_expires else {
             response.drain_body().await;
-            return Ok(false);
+            return Ok(None);
+        };
+        let Some(expire_at) = target.resolve(metadata.time_created) else {
```
lcian, line 453: Here and in the other backends too.

lcian (approved):

objectstore-service/src/backend/common.rs
```diff
@@ -80,15 +111,17 @@ pub trait Backend: fmt::Debug + Send + Sync + 'static {
...
     /// payload, and all other metadata remain unchanged.
     ///
-    /// Returns `true` when the deadline was extended or was already at least as
-    /// late as `expire_at`. Returns `false` when the object is absent, expired,
-    /// manually expired, or changed concurrently.
+    /// Returns the resolved requested deadline when the deadline was extended or
+    /// was already at least as late. Returns `None` when the object is absent,
+    /// expired, non-expiring, changed concurrently, or lacks the creation time
+    /// needed to resolve the target. The returned deadline is not necessarily the
+    /// stored deadline.
```
jan-auer, line 118: I just realized that we need to further differentiate here. We will want different status codes for missing objects and conflicts.

objectstore-service/src/backend/common.rs
```diff
@@ -324,16 +373,18 @@ pub trait HighVolumeBackend: Backend {
...
+    /// Redirects can only resolve absolute targets because tombstones do not store
+    /// creation time. Backend failures are returned as errors.
     async fn compare_and_update(
         &self,
         id: &ObjectId,
         current: Option<&ObjectId>,
         update: TieredUpdate,
         access_time: Timestamp,
-    ) -> Result<bool>;
+    ) -> Result<SetExpiryResponse>;
```
jan-auer, line 387: Note that this change is not ideal because we now have a domain-specific response in a more general API (`compare_and_update`, not `compare_and_set_expiry`). Since this is our only implementation, we can keep it this way for now, but we'll improve this in a follow-up.

objectstore-service/src/backend/gcs.rs
jan-auer: I'll record this as a dedicated follow-up; that way we can be more explicit about this.

objectstore-service/src/backend/common.rs
lcian: True. It can probably be solved by adding another layer of indirection.

lcian (approved):
Looks good to me, thanks for changing the `Option` into an `enum`.

objectstore-service/src/backend/gcs.rs
jan-auer: Filed as FS-524

PR #640, opened 2026-09-17
feat(local-fs): Support resumable uploads

lcian:
Adds support for the new Resumable Uploads API on the LocalFS backend.

Uploads use a UUIDv7 identifiers and are stored in `<root>/uploads/<upload-id>`.
To keep things as simple as possible, operations on a resumable upload mutate such file directly and use the same lock as mutations on their associated object file. This creates some possibility of contention.

Publication of a resumable upload can theoretically fail. When that happens, a subsequent offset query or put_chunk request with 0 bytes will retry publication.
It's still not clear to me if this is the behavior we want. After I start the implementation on `TieredStorage` this will likely become clearer so I'll be able to change things around.

This allows us to resolve the TODO comments in the server and Rust client, where we finally add e2e tests.


lcian:
@sentry review
bugbot run

lcian:
bugbot run

objectstore-service/src/backend/local_fs.rs
```diff
@@ -274,6 +285,105 @@ impl Backend for LocalFsBackend {
...
+        let _guard = self.locks.acquire(id).await?;
+
+        let upload_path = self.upload_path(session.upload_id);
+        let mut upload = UploadFile::open(&upload_path).await?;
+        // A previous call to `put_chunk` should have already materialized the object.
+        // If we find the object still here, it means that publication failed.
+        // Treat the session as permanently failed.
+        if upload.offset() == session.total_length.get() {
+            return Err(ErrorKind::UploadSessionGone.into());
+        }
```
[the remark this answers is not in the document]
lcian: We could make another final `put_chunk` attempt to finalize again, but ideally I'd like to avoid attempting to finalize again on an `offset` query, as that should be a read-only operation.
The question is, what should an `offset` query report if the contents are fully uploaded but we've failed to finalize the file?

objectstore-service/src/backend/local_fs.rs
```diff
@@ -663,13 +773,192 @@ impl ObjectLocks {
...
+    stored_size: u64,
+    expires_at: Option<Timestamp>,
+}
+
+impl UploadFile {
+    async fn create(path: &Path, metadata: &Metadata) -> Result<()> {
+        let mut options = OpenOptions::new();
+        options.create_new(true).read(true).write(true);
+        #[cfg(unix)]
+        options.mode(0o666);
```
matt-codecov: `Draft` also does 666 but we should probably change that

the upside of 666 is that, for this specific backend, operators can attach the fs and bring their own external solution for directory-level operations. but we should probably restrict:
- at least to 644, so nothing else can write without coordinating with our lock scheme or updating tombstones
- probably to 600, so compromising another service with the shared fs doesn't let you exfiltrate objectstore data

or we could make it a backend config with 600 as the default

objectstore-service/src/backend/local_fs.rs
```diff
@@ -663,13 +773,192 @@ impl ObjectLocks {
...
+        let payload_size = self.payload_size.checked_add(copied).ok_or_else(|| {
+            Error::new(
+                ErrorKind::BackendFailure,
+                "local-fs resumable upload size overflow",
+            )
+        })?;
+        let stored_size = self.stored_size.checked_add(copied).ok_or_else(|| {
+            Error::new(
+                ErrorKind::BackendFailure,
+                "local-fs resumable upload size overflow",
```
matt-codecov: i doubt it will ever come up but this should probably be a unique string to differentiate from the slightly different case a few lines up

objectstore-service/src/backend/local_fs.rs
```diff
@@ -663,13 +773,192 @@ impl ObjectLocks {
...
+
+    async fn publish(self, target: PathBuf) -> Result<(u64, Option<Timestamp>)> {
+        let Self {
+            file,
+            path,
+            stored_size,
+            expires_at,
+            ..
+        } = self;
+        drop(file);
```
matt-codecov, line 948: if your goal is to close the file, docs say you should probably explicitly `file.flush()` or `file.sync_all()`. it won't close right away if there are any pending IO operations.

objectstore-service/src/backend/local_fs.rs
```diff
@@ -663,13 +773,192 @@ impl ObjectLocks {
...
+        let (length, upload_id) = token
+            .split_once('.')
+            .ok_or(ErrorKind::UnknownUploadSession)?;
+        let total_length = length
+            .parse::<NonZeroU64>()
+            .map_err(|_| ErrorKind::UnknownUploadSession)?;
+        let upload_id_str = upload_id;
+        let upload_id =
+            Uuid::parse_str(upload_id_str).map_err(|_| ErrorKind::UnknownUploadSession)?;
+        if upload_id.get_version_num() != 7 || upload_id.to_string() != upload_id_str {
```
matt-codecov: this check feels unnecessary to me
- the code doesn't really rely on the UUID being specifically v7. no need to assert that
- i think we can just trust `Uuid::parse_str(s)?.to_string()` to work correctly

objectstore-service/src/backend/local_fs.rs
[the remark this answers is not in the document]
matt-codecov: this idea is a bit weird, but a zero-byte `put_chunk()` could trigger a publish retry. exceeding the max session length is already handled, so you could just get rid of this check

`upload_offset()` would stop checking for this case too. and we would want to call this out in user docs

objectstore-service/src/backend/local_fs.rs
```diff
@@ -274,6 +285,105 @@ impl Backend for LocalFsBackend {
...
+            return Err(ErrorKind::UploadSessionGone.into());
+        }
+        if content_length != 0 && offset != upload.offset() {
+            return Err(ErrorKind::UploadOffsetMismatch {
+                offset: upload.offset(),
+            }
+            .into());
+        }
+
+        let reader = StreamReader::new(stream).take(content_length);
```
matt-codecov, line 352: do we need to make sure we fully drain the stream after taking `content_length` to free up connections? or will dropping the stream do all the cleanup we might want

matt-codecov (approved):
stamp with nits, pls address before merge (or @ me for another review pass)

objectstore-service/src/backend/local_fs.rs
lcian: Using `600` now for files and `700` for dirs, and consolidated into two helpers `file_options` and `create_directories` that return/apply the correct chmod.

objectstore-service/src/backend/local_fs.rs
lcian: Yep, fair, fixed.

objectstore-service/src/backend/local_fs.rs
lcian: The assumption is that you write to the file using `append` which already performs the flushing and syncing.
It's not super clean as there is no _guarantee_ you do it, it's just something that happens to be true in today's code. `Draft` makes a similar assumption that you have to call `Draft::prepare` before `Draft::publish`.
For now, I've added comments to both, stating the assumption.

objectstore-service/src/backend/local_fs.rs
lcian: Dropping the stream drops the underlying `hyper` stream which performs a "best-effort" drain (only drains if there's still one frame left, otherwise closes the connection)
https://github.com/hyperium/hyper/blob/e3bcd379118e707b3e49185b047d49ebbaa943fc/src/proto/h1/dispatch.rs#L230-L235
https://github.com/hyperium/hyper/blob/e3bcd379118e707b3e49185b047d49ebbaa943fc/src/proto/h1/conn.rs#L848-L864

I would leave it like this, for now.
To ensure we drain, we would need a wrapper at the server level, as there are probably other scenarios where we would have to handle this. For example in GCS, the incoming stream becomes the outgoing stream to GCS. If there's an error transmitting a frame to GCS, you would ideally still want to drain the rest of the body from the client to reuse the connection.
To do this correctly you would want to do this in a "bounded" way, i.e. probably under a reasonable timeout, otherwise a stalled/slow client could keep the connection and drain task open indefinitely.

objectstore-service/src/backend/local_fs.rs
```diff
@@ -274,6 +301,113 @@ impl Backend for LocalFsBackend {
...
+        let _guard = self.locks.acquire(id).await?;
+
+        let upload_path = self.upload_path(session.upload_id);
+        let mut upload = UploadFile::open(&upload_path).await?;
+        // A previous call to `put_chunk` should have already materialized the object.
+        // If we find the object still here, it means that publication failed.
+        // Treat the session as permanently failed.
+        if upload.offset() == session.total_length.get() {
+            return Err(ErrorKind::UploadSessionGone.into());
+        }
```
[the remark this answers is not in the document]
lcian: This is a variation of the problem discussed above.

objectstore-service/src/backend/local_fs.rs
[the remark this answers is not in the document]
lcian: Technically that's what we need to do on `TieredStorage`, I think.
I've implemented it as you suggested for now. We might have to revisit later.

PR #641, opened 2026-09-18
feat(in-memory): Support Resumable Uploads

lcian:
Adds support for Resumable Uploads to the `InMemoryBackend`.
The implementation follows existing patterns and should be non-controversial.

This is useful for the upcoming implementation of Resumable on `TieredStorage`, as in that module we use the `InMemoryBackend` in both slots during tests.


lcian:
bugbot run
@sentry review

objectstore-service/src/backend/in_memory.rs
```diff
@@ -226,6 +251,102 @@ impl super::common::Backend for InMemoryBackend {
...
+    }
+
+    async fn upload_offset(&self, id: &ObjectId, token: &BackendToken) -> Result<UploadProgress> {
+        let session = self.upload_session(id, token)?;
+        let guard = session.lock().await;
+        let upload = guard.as_ref().ok_or(ErrorKind::UnknownUploadSession)?;
+        Ok(UploadProgress::Incomplete {
+            offset: upload.data.len() as u64,
+        })
+    }
```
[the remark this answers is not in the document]
lcian: The contract says that `upload_offset`:
- **Can** return `UploadProgress::Complete` repeatedly after the final chunk
- Returns `ErrorKind::UnknownUploadSession` when `token` **does not identify a known session**

In both cases here, we're not violating the contract.
The contract is intentionally vague, as the behavior might be different across backends.
After the Tiered implementation I'll see if I can tighten it and align the backends we control to all behave similarly if necessary.

objectstore-service/src/backend/in_memory.rs
```diff
@@ -226,6 +251,102 @@ impl super::common::Backend for InMemoryBackend {
...
+
+        let mut remaining = content_length;
+        while remaining != 0 {
+            let Some(bytes) = stream.try_next().await? else {
+                break;
+            };
+            let length = remaining.min(bytes.len() as u64) as usize;
+            upload.data.extend_from_slice(&bytes[..length]);
+            remaining = remaining.saturating_sub(length as u64);
+        }
```
matt-codecov, line 307: thinking adversarially, we just trust here that the user-provided content length is accurate. the stream could be much longer, and we'll consume the whole thing. our exit condition only checks if `remaining` is exactly 0

objectstore-service/src/backend/in_memory.rs
```diff
@@ -226,6 +251,102 @@ impl super::common::Backend for InMemoryBackend {
...
+
+        let offset = upload.data.len() as u64;
+        if offset != upload.total_length.get() {
+            return Ok(UploadProgress::Incomplete { offset });
+        }
+
+        let upload = guard.take().unwrap();
```
matt-codecov, line 314: is this to give up a mutable borrow on `self`?

matt-codecov (approved):
accept w a nit

objectstore-service/src/backend/in_memory.rs
lcian: Yeah but `remaining` is initially set to `content_length`, so we'll consume at most `content_length`.

objectstore-service/src/backend/in_memory.rs
lcian: We take the contents of the upload so we don't have to copy them into the final object.
Also, this way if another concurrent operation acquires the mutex, they will immediately find `None`, which is correct as this operation is completing the upload.

PR #642, opened 2026-09-21
feat(tiered): Support Resumable Uploads

lcian:
This implements Resumable Uploads in `TieredStorage`.
Notes:
- We only accept uploads bigger that exceed the `BACKEND_SIZE_THRESHOLD`, smaller ones are rejected at creation time. There was an idea to still accept the upload creation request and then force the client to upload the whole payload in a single shot by always reporting offset 0, but that cannot be achieved unless we always keep some state about which uploads exist/are active at a given moment. So, we might as well reject the creation request and then the user is still practically forced to use a regular `put` and upload in a single shot.
If we change the way we do routing, this can trivially be adapted.
- As discussed with @jan-auer, we intentionally release this while knowing it has consistency issues. I didn't see any immediate way of making this consistent, so we defer solving this problem to the Consistency project, as a proper solution will likely require some level of fundamental redesign. I've documented the two most obvious problematic scenarios for the approach I've chosen, but the list is definitely not exhaustive.

Refs FS-508


lcian:
bugbot run
@sentry review

objectstore-service/src/backend/tiered.rs
```diff
@@ -402,6 +429,189 @@ impl Backend for TieredStorage {
...
+            };
+        }
+
+        // Verify that this is indeed the final request before proceeding
+        // (i.e. that the client is not sending us a premature offset).
+        // GCS returns `Complete` when the upload was completed successfully, even if the underlying
+        // blob has been subsequently deleted, so the intention here is to prevent the creation of a
+        // dangling tombstone in such cases.
+        match self.upload_offset(id, token).await? {
+            UploadProgress::Complete => return Ok(UploadProgress::Complete),
```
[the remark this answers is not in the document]
lcian: Known issue.

objectstore-service/src/backend/tiered.rs
```diff
@@ -402,6 +429,189 @@ impl Backend for TieredStorage {
...
+                )
+                .await?;
+            return match progress {
+                UploadProgress::Incomplete { .. } => Ok(progress),
+                UploadProgress::Complete => Err(ErrorKind::UploadSessionGone.into()),
+            };
+        }
+
+        // Verify that this is indeed the final request before proceeding
+        // (i.e. that the client is not sending us a premature offset).
```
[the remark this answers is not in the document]
lcian: Known issue.

objectstore-service/src/backend/tiered.rs
```diff
@@ -402,6 +429,193 @@ impl Backend for TieredStorage {
...
+            })
+            .await?;
+        guard.advance(ChangePhase::Written);
+
+        // 4. Publish the tombstone.
+        let written = self
+            .inner
+            .high_volume
+            .compare_and_write(
+                id,
```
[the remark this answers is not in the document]
lcian: Known.

objectstore-service/src/backend/tiered.rs
```diff
@@ -392,6 +388,37 @@ impl TieredStorage {
...
+    backend: ResumableBackendChoice,
+    backend_token: BackendToken,
+    total_length: u64,
+    time_expires: Option<Timestamp>,
+}
+
+impl TieredResumableToken {
+    fn decode(id: &ObjectId, token: &BackendToken) -> Result<Self> {
+        let session: Self =
+            serde_json::from_str(token).map_err(|_| ErrorKind::UnknownUploadSession)?;
```
matt-codecov: this is weird to me. we deserialize a `BackendToken` into a type which owns a `BackendToken`. i know `BackendToken` is a type alias for `String` but perhaps we want a different type alias for whatever the inner token is vs the outer token. but maybe the trait signatures make that hard.

objectstore-service/src/backend/tiered.rs
```diff
@@ -402,6 +429,193 @@ impl Backend for TieredStorage {
...
+        let end = offset
+            .checked_add(content_length)
+            .filter(|end| *end <= session.total_length)
+            .ok_or(ErrorKind::ChunkExceedsUploadLength {
+                offset,
+                content_length,
+                upload_length: session.total_length,
+            })?;
+
+        // Non-final request; just forward the chunk.
```
matt-codecov, line 472: i don't totally get why we need special behavior.
- pass the `put_chunk()` through to GCS
- if `Incomplete`, return that, great
- if `Complete`, fetch our tiered metadata
- if tiered metadata isn't the right version, update it
- if it was already the same, log bc that's weird
- return `Complete`

would that not work? when does a `Complete` from GCS mean the ID+revision in the current request is wrong to put in Bigtable?

objectstore-service/src/backend/tiered.rs
```diff
@@ -402,6 +429,193 @@ impl Backend for TieredStorage {
...
+        // Verify that this is indeed the final request before proceeding
+        // (i.e. that the client is not sending us a premature offset).
+        // GCS returns `Complete` when the upload was completed successfully, even if the underlying
+        // blob has been subsequently deleted, so the intention here is to prevent the creation of a
+        // dangling tombstone in such cases.
+        match self.upload_offset(id, token).await? {
+            UploadProgress::Complete => return Ok(UploadProgress::Complete),
+            UploadProgress::Incomplete { offset: actual } if actual < offset => {
+                return Err(ErrorKind::UploadOffsetMismatch { offset: actual }.into());
+            }
```
matt-codecov, line 494: why only `<`? also wouldn't the GCS backend enforce this when we try to call `put_chunk`?

objectstore-service/src/backend/tiered.rs
```diff
@@ -402,6 +429,193 @@ impl Backend for TieredStorage {
...
+            };
+        }
+
+        // Verify that this is indeed the final request before proceeding
+        // (i.e. that the client is not sending us a premature offset).
+        // GCS returns `Complete` when the upload was completed successfully, even if the underlying
+        // blob has been subsequently deleted, so the intention here is to prevent the creation of a
+        // dangling tombstone in such cases.
+        match self.upload_offset(id, token).await? {
+            UploadProgress::Complete => return Ok(UploadProgress::Complete),
```
matt-codecov, line 491: i don't quite understand your comment explaining this case. could you elaborate?

my read is that, because of the `end != session.total_length` condition above, we believe the current request is the final `put_chunk()` in the object. if GCS returns `Complete` before we forward the `put_chunk()`, that means this is a re-send, which i guess means we don't know if there's a tombstone. but... i assume that means the finalized object is in GCS and we can go ahead and create the tombstone for it. what's the scenario where GCS returns `Complete` but has actually deleted the real object?

or is this meant to handle this case:
- final `put_chunk()`
- object finalized, tombstone created
- object and tombstone explicitly deleted
- stray `put_chunk()` for same object after it was deleted

objectstore-service/src/backend/tiered.rs
lcian: GCS keeps returning `Complete` after the upload has been completed successfully.
So if you did that you could have a scenario where:
1. A completes the upload
2. B deletes the object and therefore the LT revision
3. A comes in and completes the upload again; this commits a dangling tombstone.
This is the whole reason we try to commit only on the final request.
I tried to explain this in the comment below, at line 511.
The mechanism is still not perfect, as you could have 2 or more requests coming in at the same time that you all consider "final" and still end up with the same race, although at least it makes this failure harder to hit...

objectstore-service/src/backend/tiered.rs
lcian: GCS and by extension the GCS backend accept replays of a suffix of the persisted bytes.
Meaning: if current server offset is A, I can send a new chunk at offset B iff 0 <= B <= A. The only forbidden thing is B > A (hence `<` in the code).

There's no straightforward way to prevent that in the backend.
We could query the offset at every `put_chunk` to reject replays, but that also doesn't work in all cases, as you could have 2 requests coming in at the same time and observing the same offset, but then finding out that one replays a suffix of the other.

objectstore-service/src/backend/tiered.rs
lcian: Yes exactly, that's the case this is meant to handle.
As I mentioned above, even if the underlying object is deleted, then GCS's resumable API will continue to report `Complete`, as it refers to the upload, not the object.

Eventually, we might need to keep some additional state for the resumable upload in Bigtable to be able to distinguish these cases in order to achieve consistency.

objectstore-service/src/backend/tiered.rs
lcian: Fair. It's still an alias so we can create a new alias to distinguish between the two and things will just work as long as they alias the same underlying type, let me do that.

lcian: Ok I did that and simplified a bit. I think we don't need the `ResumableBackendChoice` anymore as it doesn't make sense to route resumables to `high_volume` for now.
If we later want to change this, we can do so freely, it will just invalidate the currently ongoing sessions at time of deploy.

objectstore-service/src/backend/tiered.rs
lcian: Of course, given that it's not a goal for this PR to make this consistent, we could very well remove all of these not completely effective defenses for the sake of keeping the code as simple as possible (while opening up a few more possibilities for inconsistency, but it's not like we're consistent right now).
Do you think we should do it that way?

objectstore-service/src/backend/tiered.rs
```diff
@@ -402,6 +414,178 @@ impl Backend for TieredStorage {
...
+        let session = TieredResumableToken::decode(token)?;
+        let revision = ObjectId {
+            context: id.context.clone(),
+            key: session.revision.clone(),
+        };
+        self.inner
+            .long_term
+            .upload_offset(&revision, &session.inner)
+            .await
+    }
```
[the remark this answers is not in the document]
lcian: Known issue.

PR #643, opened 2026-09-21
feat(change_stream): MultiStream impl of ChangeStream

matt-codecov:
not currently used. demo of how `ChangeStreamFactory` can be made to work with multiple `ChangeStream` impls for both storage cogs and custom garbage collection


objectstore-service/src/change_stream/factory.rs
```diff
@@ -26,6 +29,49 @@ pub enum CostTrackerConfig {
...
+        for stream in self.streams.iter() {
+            stream.update(id, expires_at);
+        }
+    }
+
+    fn delete(&self, id: &ObjectId) {
+        for stream in self.streams.iter() {
+            stream.delete(id);
+        }
+    }
```
matt-codecov, line 62: these aren't async trait methods but we might internally want to spawn them concurrently in case the underlying streams become sync by blocking on a long operation instead of returning optimistically

aldy505: WIth that being said, therefore every operation on `ChangeStream` should be blocking? Or am I misunderstanding this?

